PR #1, opened 2020-07-14
Update autoconf scripts, copy m4 macros, build instructions, and experimental Docker files.

brturn:


README.build/Ubuntu-18.04.readme
```diff
@@ -0,0 +1,47 @@
+
+------------------------------------------
+Ubuntu Server 18.04 Cloud9 bootstrap build
+------------------------------------------
+
+   Instructions to build and run burr-tools from a Cloud9 amazonlinux environment.  This will build the burrTxt2 and burrGui applications, but I have only tested burrTxt2 for functionality, as there is no GUI available.
+
+Contact: Bryan Turner (bryan.turner@pobox.com or "brturn" on GitHub)
```
akoehn, line 8: I would rather point to the github issue tracker (same for other lines like this). People are more likely to post there and others can then already find solutions to their problems.

bootstrap
```diff
@@ -1,8 +1,8 @@
 #! /bin/sh
-set -ex
+# set -ex
```
akoehn, line 2: I there a reason for disabling this? `set -e` should be in every shell script IMO to prevent a continuation after error.

bootstrap
```diff
@@ -1,8 +1,8 @@
...
 
-export DARCS_DONT_ESCAPE_ISPRINT=1
-darcs changes > ChangeLog
+# export DARCS_DONT_ESCAPE_ISPRINT=1
```
akoehn, line 4: We'll never need these again, so they can be deleted

configure.ac
```diff
@@ -68,15 +72,13 @@ case "$target_os" in
   mingw*)
     ;;
   darwin*)
-   AX_CHECK_GL
+    AX_CHECK_GL
```
akoehn, line 75: We probably don't need a darwin case if it is the same as the standard case.

PR #6, opened 2020-07-25
Windows 10 (64-bit) build instructions.  Disabled Lua script.  Restored config.h.in

brturn:
Lua script to load config file ~/.burrtools.rc was causing a seg-fault.  Haven't dug any deeper, just commented it out for now.

Bootstrap script was not creating the config.h.in file needed to complete configuration.  Restoring it for this push.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ config.log
 config.status
 doxygen.cfg
 src/config.h
-src/config.h.in
 src/stamp-h1
 src/help/helpdata.cpp
 src/burrTxt
@@ -21,4 +20,8 @@ src/burrTxt2
 src/gui/burrGui
 src/help/helpviewer
 src/unitTest
-src/config.h.in
+src/burrTxt.exe
+src/burrTxt2.exe
+src/gui/burrGui.exe
+src/help/helpviewer.exe
+src/unitTest.exe
--- a/src/flu/Flu_Tree_Browser.cpp
+++ b/src/flu/Flu_Tree_Browser.cpp
@@ -2983,7 +2983,7 @@ Flu_Tree_Browser::Node* Flu_Tree_Browser :: add_leaf( const char* path, const ch
 
 unsigned long Flu_Tree_Browser :: remove( const char *fullpath )
 {
-  return( (unsigned long)root.modify( fullpath, Node::REMOVE, rdata ) );
+  return reinterpret_cast<uintptr_t>(root.modify( fullpath, Node::REMOVE, rdata ) );
 }
 
 unsigned long Flu_Tree_Browser :: remove( const char *path, const char *text )
--- a/src/flu/Flu_Tree_Browser.h
+++ b/src/flu/Flu_Tree_Browser.h
@@ -1070,7 +1070,7 @@ class FLU_EXPORT Flu_Tree_Browser : public Fl_Group
       //! Remove the entry identified by path \b fullpath from this node
       /*! \return the unique id of the removed entry, or \c 0 if no matching entry was found */
       inline unsigned long remove( const char *fullpath )
-	{ return( (unsigned long)modify( fullpath, REMOVE, tree->rdata ) ); }
+	{ return reinterpret_cast<uintptr_t>(modify( fullpath, REMOVE, tree->rdata ) ); }
 
       //! Remove the entry identified by unique id \b id from this node
       /*! \return the unique id of the removed entry, or \c 0 if no matching entry was found */
--- a/README.build/Windows-10.readme
+++ b/README.build/Windows-10.readme
@@ -1,0 +1,89 @@
+
+------------------------------------------
+Windows 10 (64-bit) bootstrap build
+------------------------------------------
+
+   Instructions to build and run burr-tools from a Windows 10 VM.  This will build the burrTxt2 and burrGui applications.
+
+Issues/Comments?  Join us on GitHub:
+   https://github.com/burr-tools/burr-tools/issues
+
+--------------------------------
+Windows 10 Development VM
+--------------------------------
+
+Download Windows Development VM:
+    https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/
+
+Install MSYS2:
+    https://github.com/msys2/msys2-installer
+
+Use Pacman to install MingW (web page with instructions):
+    https://feaforall.com/install-c-language-gcc-compiler-windows/
+
+--------------------------------
+Downloading Packages
+--------------------------------
+
+> pacman -S base-devel gcc vim cmake
+
+  *** Add PATHs (Start -> Search -> Advanced Control Panel/System Settings -> Environment Variables -> PATH -> Add)
+    C:\msys64\mingw64\bin
+    C:\msys64\usr\bin
+
+> pacman -S 
+    zlib 
+    msys/libpng 
+    msys/zlib-devel 
+    mingw-w64-x86_64-fltk 
+    mingw-w64-x86_64-gcc
+    mingw-w64-x86_64-freeglut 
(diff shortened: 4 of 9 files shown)


.gitignore
```diff
@@ -21,4 +21,8 @@ src/burrTxt2
 src/gui/burrGui
 src/help/helpviewer
 src/unitTest
-src/config.h.in
+src/burrTxt.exe
```
akoehn, line 24: Why is config.h.in version-controlled again?  Is there a specific reason for not autogenerating it?

All the .exe files could be subsumed under `*.exe`

configure.ac
```diff
@@ -20,7 +20,13 @@ AM_INIT_AUTOMAKE([subdir-objects])
 AM_SILENT_RULES([yes])
 AM_MAINTAINER_MODE
 
-AM_CONFIG_HEADER(src/config.h)
+AC_CONFIG_HEADERS([src/config.h])
+
+AC_CHECK_HEADER([winsock2.h],
```
akoehn, line 25: Is this something we have to guard to only be used in win32 environments? I'm not familiar with autotools+windows.

src/flu/Flu_Tree_Browser.cpp
```diff
@@ -2983,7 +2983,7 @@ Flu_Tree_Browser::Node* Flu_Tree_Browser :: add_leaf( const char* path, const ch
 
 unsigned long Flu_Tree_Browser :: remove( const char *fullpath )
 {
-  return( (unsigned long)root.modify( fullpath, Node::REMOVE, rdata ) );
+  return reinterpret_cast<uintptr_t>(root.modify( fullpath, Node::REMOVE, rdata ) );
```
akoehn, line 2986: This method returns a long, but the return is cast to uintptr.  Seems wrong to me.

src/flu/Flu_Tree_Browser.h
```diff
@@ -1070,7 +1070,7 @@ class FLU_EXPORT Flu_Tree_Browser : public Fl_Group
       //! Remove the entry identified by path \b fullpath from this node
       /*! \return the unique id of the removed entry, or \c 0 if no matching entry was found */
       inline unsigned long remove( const char *fullpath )
-	{ return( (unsigned long)modify( fullpath, REMOVE, tree->rdata ) ); }
+	{ return reinterpret_cast<uintptr_t>(modify( fullpath, REMOVE, tree->rdata ) ); }
```
akoehn, line 1073: Same as above.

src/lua/luaclass.cpp
```diff
@@ -55,10 +55,16 @@ bool luaClass_c::getBool(const char *name) {
 
 /* functions to evaluate lua code */
 int luaClass_c::doFile(const char *fname) {
-  luaL_loadfile(L, fname) || lua_pcall(L, 0, 0, 0);
```
akoehn, line 58: I do not understand the reason to call lua_pcall.  Is this to extrract potential errors from the previous call?

akoehn (commented):
See comments :-)

src/lua/luaclass.cpp
akoehn: to add: Shouldn't the new code be 
```
  int ret = luaL_loadfile(L, fname);
  if (ret == 0) {
    return lua_pcall(L, 0, 0, 0);
  }
  return ret;
```

src/flu/Flu_Tree_Browser.cpp
brturn: This is a 64-bit issue; the code is using a pointer as if it is a number (eg: pointer to unsigned long).  This is no longer valid in general (ptr size not matching size of a long).  The reinterpret cast changes the pointer value into a number (doing whatever is necessary for the platform at hand; uintptr is the magic type for this), then C++ auto-converts from that number to an unsigned long.

src/lua/luaclass.cpp
brturn: This one is rather odd.  So the function returns an int, but has no return statement, which if I understand the semantics correctly means the "truthy" output of void, which is zero in integer space.  As for the Lua calls, the load file prepares the Lua VM for operation, loading and registering functions, etc.  The "pcall" actually executes the Lua VM.  Honestly don't care one way or another which return value to use.

.gitignore
brturn: It's a duplicate line, it's still git-ignored.

configure.ac
brturn: It works on my Mac and Linux build systems.  It is just checking for the location of a header, and if found, setting some environment variables.  In the case of non-windows systems, it does not find the header so the vars are empty.

src/flu/Flu_Tree_Browser.cpp
akoehn: I understand the idea, but the code not casts long to uintptr and then -- if I understand it correctly -- the return value is cast to long again because the method returns long.  In the best case, this seems to be a noop.

On windows, size(long) < size(ptr), so there is a problem with addressability somewhere already and that underlying problem should be rather fixed.

In addition, this code does not compile on my debian with gcc 10 because `cstdint` is not included.

src/lua/luaclass.cpp
akoehn: > So the function returns an int, but has no return statement, which if I understand the semantics correctly

According to the c++ standard, this is undefined behavior.  Therefore, I would return the lua return value.

src/flu/Flu_Tree_Browser.cpp
brturn: Did some more digging.  The issue is that Node::modify returns a Node*, but the value is not always a pointer.  You can see this, for example around line 3353 the code returns the actual pointer to a Node, while in the same function, around line 3506 it returns the ID (an int) cast to a pointer.  The remove() function is documented in the header file to "return the unique ID of the node removed or 0 if none was found".  So, the reinterpret cast in this context is correct (pointer is converted to a number, zero is left alone).

akoehn: The point I was trying to make is that this method returns a long (line 2984) and therefore the cast inside the method is pointless because it is then cast to long again upon return

akoehn: Sorry its late here, will  reply with more sleep

brturn: I couldn't remember why I used the reinterpret cast in the first place, so I checked in a hard-cast which compiles fine on my Mac.  Then I remembered that this is the Windows branch.. so back to Windows and MINGW... This is the error it is giving me with a hard-cast.  Reinterpret-cast compiles cleanly on all platforms for me (Mac, Ubuntu, Windows).

gui/../flu/Flu_Tree_Browser.h:1075:12: error: cast from 'Flu_Tree_Browser::Node*' to 'long unsigned int' loses precision [-fpermissive]
 1075 |     return (unsigned long)ret;
      |            ^~~~~~~~~~~~~~~~~~

akoehn: Let me rephrase my critique: It does not matter what kind of casting is going on inside this method because the method itself returns a value of type double.  All the casts are doing is quieting the compiler.

Under Linux and Mac, long and pointer have the same size, so all the casts are noops. Under Win64, a long is only 32 bit but a pointer is 64 bit. The method returns a long, i.e. 32 bit. Converting a 32 bit double to uintptr (i.e. 64 bit) and then have the compiler add an implicit cast to double again (i.e. 32 bit) does not do anything meaningful as well.

The problem here are the return types that would need to be changed.

We can still merge this as it creates no new harm, but in that case we should have a TODO somewhere.

PR #7, opened 2020-10-27
Correct Lua return values, clean up all compiler warnings in BT code

akoehn:
There are still warnings in third-party code (mostly FLU)

Make Flu_Tree_Browser::Node::_id an uintptr_t, use uintptr_t in
Flu_Tree_Browser::{remove, modify} as modify can either return an id
or a pointer

rotater_c: Add virtual destructor

Catch exceptions by reference instead of by value

remove unused variables

Reformat lines with for and while loops to make them more legible

mark all fall throughs in switch statements

Don't assert where we always need to check

xmlWriter_c destructor: mark that it can throw exceptions


akoehn:
@brturn can you comment and let me do the merge? I want to clean up the commits; those should have been one.

brturn (approved):
Changes look good.  Have not tried to compile on Mac/Win/Linux though, but I'll check soon.

PR #9, opened 2021-02-23
retina fix

smiteo:
this may not be the most elegant way to fix the retina issue...
but the  issue is with FLTK and the way it handles retina displays.

the xml changes are just so I could get it to compile

### DIFF
--- a/src/tools/xml.cpp
+++ b/src/tools/xml.cpp
@@ -54,7 +54,7 @@ xmlWriter_c::xmlWriter_c(std::ostream & str) : stream(str), state(StBase)
   stream << "<?xml version=\"1.0\"?>\n";
 }
 
-xmlWriter_c::~xmlWriter_c(void) noexcept(false)
+xmlWriter_c::~xmlWriter_c(void)// noexcept(false)
 {
   if (tagStack.size() > 0)
     throw xmlWriterException_c("Not all tags were closed");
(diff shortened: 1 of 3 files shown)


tomburns (approved):

tomburns:
@brturn this looks good to me, should we go ahead and merge it to try and get a fixed mac build out?

akoehn:
I'm pretty sure the `noexcept false` was there for a reason. Iirc the deconstructor can throw an exception which has to be declared as it can have interesting side effects.

I would also like to try this change on a different platform veggie merging.

src/tools/xml.cpp
```diff
@@ -54,7 +54,7 @@ xmlWriter_c::xmlWriter_c(std::ostream & str) : stream(str), state(StBase)
   stream << "<?xml version=\"1.0\"?>\n";
 }
 
-xmlWriter_c::~xmlWriter_c(void) noexcept(false)
```
akoehn, line 57: This change seems to be for an ancient C++ version; I would like to keep the code clean wrt at least c++11.
```
tools/xml.cpp: In destructor ‘xmlWriter_c::~xmlWriter_c()’:
tools/xml.cpp:60:5: warning: ‘throw’ will always call ‘terminate’ [-Wterminate]
   60 |     throw xmlWriterException_c("Not all tags were closed");
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tools/xml.cpp:60:5: note: in C++11 destructors default to ‘noexcept’
```

akoehn (changes requested):

akoehn:
Otherwise, tested on Linux and everything still seems to work.

PR #10, opened 2021-02-26
GitHub ci

akoehn:
I think we should be fairly strict regarding code quality because it is hard to fix stuff later on in a distributed setting like we have.

This PR contains:
 - -Werror for the burrtools code (but not for fltk etc)
 - a short message to the AUTHORS file
 - GitHub CI configuration


smiteo (approved):

akoehn:
manually merged.

PR #15, opened 2022-06-17
Remove 'register' keyword which was removed in C++17

### LINKED ISSUE #14: Does not compile with C++ 17 compiler
The code does not compile with a compiler in C++ 17 mode, such as gcc/cpp 11.3.

```
lib/assembler_0.h:94:40: error: ISO C++17 does not allow ‘register’ storage class specifier [-Werror=register]
   94 |   void cover_row(register unsigned int r);
      |                                        ^
```

The `register` keyword was deprecated and finally removed in C++ 17.

raoulb:
Fixes #14

### DIFF
--- a/src/lib/assembler_0.cpp
+++ b/src/lib/assembler_0.cpp
@@ -918,12 +918,12 @@ void assembler_0_c::uncover(unsigned int col) {
 /* remove all the columns from the matrix in which the given
  * row contains ones
  */
-void assembler_0_c::cover_row(register unsigned int r) {
+void assembler_0_c::cover_row(unsigned int r) {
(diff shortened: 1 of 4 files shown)


TheRammer (approved):
I independently resolved these same warnings ("ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]") and resulted in the same diffs.

raoulb:
Someone with the suitable privileges should merge it.

akoehn:
My pleasure.  Somehow lost this PR in my ~16k unread emails :-/

raoulb:
Thanks :)

PR #19, opened 2024-08-26
Fix std iterator

raoulb:
Handle deprecation of std::iterator in C++17


akoehn:
whole code base moved to c++20 now (and meson) so this is not applicable anymore.

PR #20, opened 2024-08-26
Fix dangling pointer warning

raoulb:
Fix warning about possibly dangling pointers.


akoehn:
Sorry, did not write here earlier: we switched over to meson some time ago and this PR is not applicable anymore.

Thanks for the PR though, appreciated to see people caring about code quality :)

PR #21, opened 2024-08-27
Fix overloaded virtual

raoulb:
Try to fix error about overloading virtual methods. This patch makes the code compile again and should not have serious drawbacks, but it's not a perfect fix either.

This might be related to some GCC issues:

  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=20423
  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109740


ZacharyTalis:
This PR allows me to compile and run BurrTools on my machine, whereas I couldn't before! <33333

On `make install` I get an error that I don't know how I'd fix: `/usr/bin/install: will not overwrite just-created '/usr/local/share/doc/burrtools/README' with 'README'`

But I'm able to run `burrGui` directly, so this isn't a showstopper :3

(system info: arch linux / intel cpu)

raoulb:
PR #22 should fix the readme overwrite issue.

akoehn:
... sorry for taking two years and three days for this PR.

PR #24, opened 2024-09-19
WiP / Do Not Merge: Disable picking

raoulb:


akoehn:
@raoulb I lost track, is this something you still want to implement somewhere? If so, I would propose to use a new branch based on current main.

PR #25, opened 2025-09-08
Prevent click event every movement in 3d view

mbrown1413:
This solves the issue of clicking (with ctrl, shift or alt) to add or remove voxels, but accidentally moving the mouse slightly while the mouse button is held down, causing way more voxels to be added or removed then intended. Every mouse move event would trigger another event, which quickly does something the user likely didn't mean to.

To reproduce: hold down shift and the left mouse button in the 3D view. Drag the mouse over a voxel. A stack of voxels will rapidly be added toward the camera.

This was originally intended behavior (see 8782ebd30cc99323257ff4fe5702b0da56d2986c) meant to make it "much more easy to 'drill' holes". However, I think it causes a bigger problem than it solves. The use case of rapidly adding or removing voxels isn't that common and it's not a controlled way to do it anyways. Doing this on accident is much too easy.


smiteo:
I think a config option is the best bet.  Let the crazy people use the old behavior if they want.

PR #26, opened 2025-09-12
Fix check-build GitHub workflow

mbrown1413:
Fixes various issues preventing the check-build GitHub action from completing successfully. The individual commits have explanations of each fix.

A passing build can be found here: https://github.com/mbrown1413/burr-tools/actions/runs/17569388721


akoehn:
Thanks, especially for the good commit messages, that was very helpful for reviewing :)

PR #27, opened 2025-09-25
Undo / redo functionality

mbrown1413:
This implements undo and redo by saving the full state of the puzzle after each change. Under the file menu is now an option to undo and redo, with keyboard shortcuts Ctrl+Z and Ctrl+Y.

Undo history is saved in XML format for now, since that was easiest way to make the implementation obviously correct. We may need to do something more intelligent in the future, as this takes more time and space than simply copying the data structure, even though that would be more difficult.

In order to avoid thread safety issues, undo and redo is disabled while the solve thread is running. Undo states are also not recorded while the solve thread is running. This means that performing an undo after a solve finishes will undo all changes performed during the solve. This isn't usually a problem since most ways to edit the puzzle are disabled while solving. See the comment in undomanager.h for a full discussion of handling thread safety.

Along with the actual puzzle state, we also store a message about what edit was made which is displayed later when the user performs an undo or redo.

There is a small second commit which fixes an issue where keyboard shortcuts, like "F3" for loading a new puzzle, would not work while on the "Puzzle" activity tab.

----

Because this is a large enough change, before this merges I'd like some beta testing testing and input on some of the design decisions I've made. Specifically:
* Does this work well with people's usual workflow?
* Does storing undo states as XML take up too much memory for large puzzles?
* Does storing undo states take too long for large puzzles?

The largest puzzle I have 67 MB. It takes ~0.5 seconds to save, and (if you don't modify any pieces used, which clears solutions) each undo state takes up the same 67 MB. I feel like this is a pretty rare case though. I can think of a few options for addressing speed and memory concerns:

* Delete the oldest undo state when the total size exceeds a threshold
* Copy the puzzle directly in memory instead of converting to and from XML
* Add an option to disable undo
* Compress the data while stored
* Consider it a rare case that doesn't need to be fixed
* Consider it good enough for now but work on one of the above in the future (I'm of course willing to do this)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -246,6 +246,7 @@ gui_burrGui_SOURCES += gui/statuswindow.cpp gui/statuswindow.h
 gui_burrGui_SOURCES += gui/stlexport.cpp gui/stlexport.h
 gui_burrGui_SOURCES += gui/togglebutton.cpp gui/togglebutton.h
 gui_burrGui_SOURCES += gui/tooltabs.cpp gui/tooltabs.h
+gui_burrGui_SOURCES += gui/undomanager.cpp gui/undomanager.h
 gui_burrGui_SOURCES += gui/vectorexportwindow.cpp gui/vectorexportwindow.h
 gui_burrGui_SOURCES += gui/view3dgroup.cpp gui/view3dgroup.h
 gui_burrGui_SOURCES += gui/voxelframe.cpp gui/voxelframe.h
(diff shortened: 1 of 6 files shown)


mbrown1413:
Rebased onto master and build passes now!

hypnoz:
I created a test version of burr-tools with undo/redo functionality here if you want to test it out:

https://github.com/hypnoz/burr-tools/

grssam:
Do we really want to store the entire change history in XML. i.e. do we need persistent undo across file reopens?

I'd say the most critical thing is in-memory undo/redo and upto a fixed stack size (say go back upto 50 actions) - including things like adding removing piece, voxel, puzzle problem, piece in a problem etc.

Making puzzle file from few KB to 50+MB should be avoided at all costs and that's probably not even the 90th percentile worst case scenario! People make A LOT of changes in their BT file until its in its final form.

PR #38, opened 2026-08-16
Add STL export of an assembled solution

bramcohen:
Adds a second mode to the STL export dialog: instead of a single shape you can export all pieces of a saved solution in their assembled arrangement. Each piece is meshed individually with the regular exporter parameters (offset, bevel, etc.), so the result is one STL file containing separate solids for the pieces, positioned as assembled, with the gaps the offset produces between them. Useful for printing or displaying an assembled model, or importing pre-arranged pieces into a slicer.

The dialog gains radio buttons for **single shape** / **assembled solution**, a problem selector and a solution number input; all exporter parameters stay adjustable and the 3D preview shows the assembled result live, just like the single-shape preview.

Implementation:

- `assembly_c::createPieceSpaces()` creates one voxel space per placed piece of an assembly (same placement logic as `createSpace()`). All spaces get the size of the whole assembly, so the pieces keep their relative placement — which makes the mesh generators of every grid type position the pieces correctly without any mesh-level transformations.
- `stlExporter_c::write()` gets an overload taking several shapes that writes them into a single STL file (the existing single-shape `write` now delegates to it).
- The dialog preview merges the per-piece meshes into one polyhedron for display and reports piece count and total volume.

**Update:** the solution tab also gets an **STL** button (next to the disassembly buttons) that opens this dialog directly in assembled-solution mode with the currently viewed problem and solution preselected. Active whenever a solution is shown, the grid supports STL export, and the solver is not running.

Two bug fixes ride along because the feature exposed them:

- `Polyhedron::finalize()` dereferences `connections.end()` when the vertex pair being matched is the last key in the multimap (both counting and angle-sorting loops). This pre-existing UB in the shared STL meshing path crashed under ASan the moment several meshes were generated back to back. (Same fix as in #37; whichever lands first, the other rebases cleanly.)
- `assembly_c::createSpace()` wrapped the `pc->transform(...)` call in plain `bt_assert`, which compiles to *nothing* under NDEBUG — so in `-Db_ndebug=true` builds, "Import Assms" placed every piece untransformed. Now uses `bt_assert2`, which keeps evaluating its expression.

Verified headlessly on every example puzzle (all grid types — cubes, spheres, rhombic, tetra-octa, triangular prisms): for every saved solution, the per-piece spaces have zero overlaps and their union is voxel-identical to `createSpace()`'s merged result; STL export of each first solution succeeds and renders correctly (assembled model with per-piece seams), all ASan-clean.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -1721,6 +1721,17 @@ void mainWindow_c::cb_STLExport(void) {
   }
 }
 
+static void cb_SolutionStlExport_stub(Fl_Widget* /*o*/, void* v) { ((mainWindow_c*)v)->cb_SolutionStlExport(); }
+void mainWindow_c::cb_SolutionStlExport(void) {
+  stlExport_c w(puzzle);
+  w.selectSolution(solutionProblem->getSelection(), (unsigned int)SolutionSel->value()-1);
+  w.show();
+
+  while (w.visible()) {
+    Fl::wait();
+  }
+}
+
 static void cb_StatusWindow_stub(Fl_Widget* /*o*/, void* v) { ((mainWindow_c*)v)->cb_StatusWindow(); }
 void mainWindow_c::cb_StatusWindow(void) {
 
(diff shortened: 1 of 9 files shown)


tomburns:
@bramcohen conflicts here but lgtm

PR #41, opened 2026-08-16
Use the native file dialog on macOS

bramcohen:
All file selection (Load Puzzle, Save Puzzle As, Import PuzzleSolver3D, STL export, vector image export) went through the old-style FLTK file chooser, which looks quite foreign on macOS.

A small wrapper (`fileChooser` in `src/gui/filechooser.{h,cpp}`) now routes these through `Fl_Native_File_Chooser` on macOS, which shows the modern system open/save panels — with the proper dialog type per call site (open vs. save), the file-type filter, and the preselected directory/file name where one exists. On all other platforms the wrapper falls straight through to `fl_file_chooser`, so nothing changes there.

FLTK's macOS backend for the native chooser references the `UniformTypeIdentifiers` framework (macOS 11+, matching the existing deployment target), so that framework is added to the linked Apple frameworks in `meson.build`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -142,6 +142,7 @@ gui_src = [
   'src/gui/separator.cpp',
   'src/gui/statusline.cpp',
   'src/gui/statuswindow.cpp',
+  'src/gui/filechooser.cpp',
   'src/gui/stlexport.cpp',
   'src/gui/togglebutton.cpp',
   'src/gui/tooltabs.cpp',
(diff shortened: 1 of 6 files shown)


akoehn:
is there a reason to only use native file choosers on macos?

bramcohen:
> is there a reason to only use native file choosers on macos?

This PR is so far intentionally narrow because I'm on a mac and can't readily test other platforms. It's probably a good idea to use the native file open everywhere

akoehn:
Tested it on Linux and windows (via wine) and native file chooser works on all platforms so I extended the PR.

akoehn (approved):

PR #42, opened 2026-08-16
Don't let STL groove filling add material at concave corners

bramcohen:
The STL export's groove filling (`fillPolyhedronHoles`) replaces the bevel/offset faces between two real faces with a single spanning face. Between coplanar faces and at convex corners that's correct — but when the two real faces meet in a **concave** corner, the replacement face bridges straight across the corner, adding a wedge of material that lies outside the voxel shape. The seam capping (`findOptimizedFaces`) does the same when closing rings around such corners.

The practical consequence: wherever two exported pieces meet in a step — e.g. the two halves of a flat crossing where one piece passes over the other — the phantom wedges make the exported STLs **physically overlap** by up to ~half a millimetre per crossing corner. Found while building a woven-band model whose pieces stack this way at every crossing; slicers doing boolean logic on the pieces get visibly confused by the overlaps.

**Fix:** a snapshot of the original (pre-fill) surface is taken at entry, and both the groove replacements and the seam caps are validated against it: probe points just below the candidate face's surface (centre + interior points toward each corner, at a shallow scale-relative depth) must lie inside the solid (ray-parity test). If the check fails, the groove faces are kept as-is (for replacements) or the capping falls back to the conservative triangle fan (for seams). Concave edges thereby keep their bevel groove instead of being bridged across.

**Verified** (headless harness diffing fixed vs. unfixed exports of every shape of every example puzzle plus a large rhombic model, 53 meshes):
- all exports remain watertight;
- no export *gains* volume (the fix can only remove the phantom wedges — up to 0.5% on strongly concave shapes, 0.003% on typical pieces);
- a minimal two-piece crossing repro goes from 0.27mm³ interpenetration to exactly zero, with all six pieces of the real model now exporting with clean pairwise clearance;
- visual diff of exports is imperceptible (hairline grooves at concave edges instead of bridged chamfers);
- export time unchanged in practice (~0.9s for a 6-piece rhombic model including the probing).

The `Polyhedron::finalize()` past-the-end fix rides along as the first commit (same as in #37–#40; it dedupes on merge, and the verification harness trips over it without the fix).

### DIFF
--- a/src/halfedge/modifiers.cpp
+++ b/src/halfedge/modifiers.cpp
@@ -29,6 +29,112 @@ using namespace std;
 
 const float Epsilon=1.0e-5;
 
+/* a snapshot of the faces of a polyhedron (as triangle fans) used to test
+ * whether a point lies inside the original solid */
+typedef vector<float> triSnapshot_c;   // 9 floats per triangle
+
+static void snapshotFaces(const Polyhedron & poly, triSnapshot_c & snap)
+{
+  for (Polyhedron::const_face_iterator it = poly.fBegin(); it != poly.fEnd(); ++it)
+  {
+    const Face * f = *it;
+    if (f->hole()) continue;
+
+    Face::const_edge_circulator e = f->begin();
+    Face::const_edge_circulator sentinel = e;
+    e++;
+    Vector3Df start = (*e)->dst()->position();
+    e++;
+    do
+    {
+      const Vector3Df & b = (*e)->dst()->position();
+      e++;
+      const Vector3Df & c = (*e)->dst()->position();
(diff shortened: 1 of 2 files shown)


bramcohen:
Follow-up commit: the original probe-based validation turned out to reject legitimate groove fills phase-dependently along bevelled lattice edges, leaving grooves open at alternating cells (pitted-looking edges). Replaced with an exact local concavity test for groove quads, and depth-tolerant probing for seam caps. Verified: constant corner chamfers matching master's clean output, still zero pairwise piece intersections (master: 11 intersecting pairs up to 0.57mm³ on the same model), and bevel-0 exports geometrically identical to master.

PR #44, opened 2026-09-05
Gui ergonomics

akoehn:
selection of ported UI adjustments from https://github.com/hypnoz/burr-tools

The code looks good to me but I did not have time to validate functionality. You can download binaries for your platform and check changes and report back.

### DIFF
--- a/src/gui/blocklistgroup.cpp
+++ b/src/gui/blocklistgroup.cpp
@@ -22,6 +22,9 @@
 
 #include "BlockList.h"
 
+#include <FL/Fl.H>
+#include <FL/Fl_Slider.H>
+
 // some tool widgets, that may be swapped out later into another file
 
 static void cb_BlockListGroupList_stub(Fl_Widget* o, void* /*v*/) { ((LBlockListGroup_c*)(o->parent()))->cb_list(); }
@@ -43,6 +46,21 @@ void LBlockListGroup_c::cb_list(void) {
 
 static void cb_BlockListGroupSlider_stub(Fl_Widget* o, void* /*v*/) { ((LBlockListGroup_c*)(o->parent()))->cb_slider(); }
 void LBlockListGroup_c::cb_slider(void) { List->setShift((int)Slider->value()); }
+
+int LBlockListGroup_c::handle(int event) {
+
+  if (event == FL_MOUSEWHEEL) {
+    double max = Slider->maximum();
+    double v = Slider->value() + Fl::event_dy() * 24.0;
+    if (v < 0) v = 0;
+    if (v > max) v = max;
+    Slider->value(v);
+    List->setShift((int)v);
+    return 1;
+  }
+
+  return Fl_Group::handle(event);
+}
 
 LBlockListGroup_c::LBlockListGroup_c(int x, int y, int w, int h, BlockList * l) : Fl_Group(0, 0, 100, 100), layoutable_c(x, y, w, h), List(l) {
 
--- a/src/gui/blocklistgroup.cpp
+++ b/src/gui/blocklistgroup.cpp
@@ -50,8 +50,15 @@ void LBlockListGroup_c::cb_slider(void) { List->setShift((int)Slider->value()); 
 int LBlockListGroup_c::handle(int event) {
 
   if (event == FL_MOUSEWHEEL) {
+    if (!Fl::event_inside(this))
+      return 0;
+
+    int dy = Fl::event_dy();
     double max = Slider->maximum();
-    double v = Slider->value() + Fl::event_dy() * 24.0;
+    if (dy == 0 || max <= 0)
+      return 0;
+
+    double v = Slider->value() + dy * 24.0;
     if (v < 0) v = 0;
     if (v > max) v = max;
     Slider->value(v);
--- a/src/gui/configuration.cpp
+++ b/src/gui/configuration.cpp
@@ -106,7 +106,8 @@ void configuration_c::parse() {
   }
 }
 
-void configuration_c::register_entry(const char *cnf_name, cnf_type cnf_typ, void *cnf_var, long maxlen, bool dialog, const char * dtext, const char * def) {
+void configuration_c::register_entry(const char *cnf_name, cnf_type cnf_typ, void *cnf_var, long maxlen, bool dialog, const char * dtext, const char * dhelp, const char * def) {
+
   config_data *t = new config_data;
 
   t->next = first_data;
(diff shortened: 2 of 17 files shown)


src/gui/blocklistgroup.cpp
```diff
@@ -44,6 +47,21 @@ void LBlockListGroup_c::cb_list(void) {
 static void cb_BlockListGroupSlider_stub(Fl_Widget* o, void* /*v*/) { ((LBlockListGroup_c*)(o->parent()))->cb_slider(); }
 void LBlockListGroup_c::cb_slider(void) { List->setShift((int)Slider->value()); }
 
+int LBlockListGroup_c::handle(int event) {
```
tomburns, line 50: **medium** — missing `Fl::event_inside(this)` guard.

This handler returns 1 for every `FL_MOUSEWHEEL` without the guard this PR adds to `LView3dGroup::handle`, so it becomes a window-wide wheel sink (see the two-pass dispatch described in the review summary).

Repro: on the Shape tab, hover the voxel grid editor and scroll. `gridEditor_c::handle` returns 0 for the wheel, so the second pass hands the event to a child *not* under the cursor and the piece-selector list scrolls instead. It's more visible on the Problems tab, which has four `LBlockListGroup_c`/`LConstraintsGroup_c` instances live at once: scrolling anywhere that doesn't itself consume the wheel scrolls whichever list the traversal reaches first, not the one under the pointer.

Two smaller issues in the same handler:
- it swallows the event when there is nothing to scroll (`Slider->maximum()` is still 0 before the first `RS_CHANGEDHIGHT` callback);
- it returns 1 after doing nothing for horizontal trackpad swipes (`event_dy() == 0`).

tomburns (commented):
Reviewed `origin/master...origin/gui-ergonomics`. Nice set of ergonomics fixes — the `Fl::event_inside(this)` guard added to `LView3dGroup::handle` is exactly the right shape. Four inline findings below, two of which I'd fix before merge.

**The mechanism the wheel findings hinge on:** `Fl::handle_` sends `FL_MOUSEWHEEL` to the *window*, and `Fl_Group::handle`'s wheel case makes **two passes** — first over children under the cursor, then over children *not* under the cursor (verified against `release-1.4.4`, the version pinned in `subprojects/fltk.wrap`). So any widget that unconditionally returns 1 for the wheel becomes a window-wide wheel sink whenever the widget genuinely under the pointer declines the event.

### Checked and cleared

- `LView3dGroup::handle` returning `Fl_Group::handle(event)` instead of a hardcoded 0 — the side effects were already happening (the base handler was called and its result discarded), so only the return value changes, and correctly: `Fl::pushed()`/`Fl::belowmouse()` now settle on the voxel frame instead of falling through to the parent.
- `configuration.cpp` layout rework — the `LFl_Box(0, 0)` two-int calls resolve unambiguously to the non-label ctor, `fl_measure(NULL, ...)` is safe in FLTK, grid rows/columns are consistent, `body`/`helpRow`/`btnRow` are all `end()`-ed, and the weight-distribution loop in `layouter_c::calcLayout` cannot divide by zero for these column/weight combinations.
- `t->widget` is uninitialized in `register_entry`, but `restoreDialogDefaults` is only reachable from the dialog's own button and `dialog()` assigns the pointer for every `CT_BOOL` entry before `show()` — no reachable failure.
- New `reversescrollzoom` entry — correctly defaulted in `parse()`, persisted by the destructor, read at event time, so no static-init-order concern.
- `StatPieceInfo` overload — `char txt[160]` with `snprintf(txt, sizeof(txt), ...)` cannot overflow; `getCursorX/Y/Z` are clipped to the shape bounds by `gridEditor_c::handle`; `pc` is range-checked.
- Doubled grid-editor cell caps (20→40, 30→60, 40→80) — `calcParameters` still derives the size from available space and takes the minimum, the parity constraints (`sx & 1`, `&= ~1`) still hold, and `calcGridPosition` shares the same routine, so hit-testing stays consistent.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

src/gui/constraintsgroup.cpp
```diff
@@ -42,6 +45,21 @@ void LConstraintsGroup_c::cb_list(void) {
 static void cb_ConstraintsGroupSlider_stub(Fl_Widget* o, void* /*v*/) { ((LConstraintsGroup_c*)(o->parent()))->cb_slider(); }
 void LConstraintsGroup_c::cb_slider(void) { List->setShift((int)Slider->value()); }
 
+int LConstraintsGroup_c::handle(int event) {
```
tomburns, line 48: **medium** — identical missing `Fl::event_inside(this)` guard, same consequence as in `blocklistgroup.cpp`.

Worth noting the knock-on effect here: because this handler intercepts the wheel *before* delegating to `Fl_Group::handle`, the companion fix in `BlockList.cpp:811` (`ColorConstraintsEdit::handle` returning 0 for `FL_MOUSEWHEEL`) can never be reached through this group — the group is now the sink that `ColorConstraintsEdit` used to be.

src/gui/view3dgroup.cpp
```diff
@@ -69,17 +70,19 @@ LView3dGroup::LView3dGroup(int x, int y, int w, int h) : Fl_Group(0, 0, 50, 50),
...
-  switch(event)
-  {
-    case FL_MOUSEWHEEL:
-      slider->value(slider->value() + 0.1*Fl::e_dy);
-      View3D->setSize(exp(6-slider->value()));
-      return 1;
+    int dy = Fl::event_dy();
+    if (config.reverseScrollZoom())
+      dy = -dy;
+    slider->value(slider->value() + 0.1 * dy);
```
tomburns: **low** — the new value isn't clamped to the slider's `minimum()`/`maximum()` (0..6), and `Fl_Valuator::value()` does not clamp on its own.

Scrolling past either end keeps accumulating: the knob sticks at the end while the internal value drifts, `exp(6 - value)` grows unboundedly, and the user has to scroll back the same number of notches before the zoom visibly responds.

The behaviour predates this PR, but this line is rewritten here and the two *new* wheel handlers added in the same change do clamp — so it seems worth closing the inconsistency now.

src/gui/configuration.cpp
```diff
@@ -220,20 +305,39 @@ void configuration_c::dialog(void) {
...
   btn->callback(cb_ConfigDialog_stub, win);
-  btn->pitch(5);
+  btn->setMinimumSize(bw, th + 10);
+
+  btnRow->end();
+  body->end();
 
   win->end();
-  win->label("Configuration");
+  win->label("Settings");
```
tomburns, line 340: **low, cosmetic** — the dialog title is renamed to `"Settings"`, but the menu entry that opens it is still `"Config"` (`mainwindow.cpp:2153`). Either rename the menu item to match or keep the title as `"Configuration"`.

src/gui/blocklistgroup.cpp
```diff
@@ -44,6 +47,21 @@ void LBlockListGroup_c::cb_list(void) {
...
+
+  if (event == FL_MOUSEWHEEL) {
+    double max = Slider->maximum();
+    double v = Slider->value() + Fl::event_dy() * 24.0;
+    if (v < 0) v = 0;
+    if (v > max) v = max;
+    Slider->value(v);
+    List->setShift((int)v);
+    return 1;
+  }
```
tomburns, line 67: Suggested fix for the wheel-sink finding. Three changes, all of which hand the event back to FLTK instead of swallowing it:

- `Fl::event_inside(this)` — matches the guard already added to `LView3dGroup::handle`, so this group stops consuming wheel events aimed at a sibling.
- `max <= 0` — nothing to scroll yet (before the first `RS_CHANGEDHIGHT` callback).
- `dy == 0` — horizontal trackpad swipe; previously returned 1 after doing nothing.

`Fl::event_dy()` is hoisted into `dy` only to avoid calling it twice.

```suggestion
  if (event == FL_MOUSEWHEEL) {
    if (!Fl::event_inside(this))
      return 0;

    int dy = Fl::event_dy();
    double max = Slider->maximum();
    if (dy == 0 || max <= 0)
      return 0;

    double v = Slider->value() + dy * 24.0;
    if (v < 0) v = 0;
    if (v > max) v = max;
    Slider->value(v);
    List->setShift((int)v);
    return 1;
  }
```

src/gui/constraintsgroup.cpp
```diff
@@ -42,6 +45,21 @@ void LConstraintsGroup_c::cb_list(void) {
...
+
+  if (event == FL_MOUSEWHEEL) {
+    double max = Slider->maximum();
+    double v = Slider->value() + Fl::event_dy() * 24.0;
+    if (v < 0) v = 0;
+    if (v > max) v = max;
+    Slider->value(v);
+    List->setShift((int)v);
+    return 1;
+  }
```
tomburns, line 65: Same fix as in `blocklistgroup.cpp` — identical handler, identical failure mode.

Worth applying both together: with only one of them guarded, the other still acts as the sink on the Problems tab, where four of these groups are live at once.

```suggestion
  if (event == FL_MOUSEWHEEL) {
    if (!Fl::event_inside(this))
      return 0;

    int dy = Fl::event_dy();
    double max = Slider->maximum();
    if (dy == 0 || max <= 0)
      return 0;

    double v = Slider->value() + dy * 24.0;
    if (v < 0) v = 0;
    if (v > max) v = max;
    Slider->value(v);
    List->setShift((int)v);
    return 1;
  }
```

src/gui/view3dgroup.cpp
```diff
@@ -69,17 +70,19 @@ LView3dGroup::LView3dGroup(int x, int y, int w, int h) : Fl_Group(0, 0, 50, 50),
...
+    View3D->setSize(exp(6 - slider->value()));
```
tomburns: Suggested fix for the unclamped zoom accumulation. Clamps against the slider's own `minimum()`/`maximum()` rather than hardcoding 0..6, and feeds the clamped `v` to `setSize` so the two can't drift apart.

```suggestion
    double v = slider->value() + 0.1 * dy;
    if (v < slider->minimum()) v = slider->minimum();
    if (v > slider->maximum()) v = slider->maximum();
    slider->value(v);
    View3D->setSize(exp(6 - v));
```

src/gui/configuration.cpp
tomburns, line 340: No committable suggestion for this one — the fix belongs on the menu side, at `mainwindow.cpp:2170` (not 2153, as I wrote above), which isn't part of this diff. The one-line change, if you want the menu to match the new dialog title:

```diff
-  {"Config",           0, cb_Config_stub,      0, 0, 0, 0, 14, 56},
+  {"Settings",         0, cb_Config_stub,      0, 0, 0, 0, 14, 56},
```

Equally fine to leave the menu alone and revert this line to `"Configuration"` — the point is just that the two currently disagree.

tomburns (commented):
Follow-up to my review above: here are committable suggestions for three of the four findings.

One correction to that review — I cited the menu entry as `mainwindow.cpp:2153`; on this branch it's actually **`mainwindow.cpp:2170`**. That line sits outside the diff, so GitHub won't let me anchor a committable suggestion to it; the one-liner is in the fourth comment below.

🤖 Suggestions generated with [Claude Code](https://claude.com/claude-code)

src/gui/blocklistgroup.cpp
tomburns: Applied in e48900d (batched with the `constraintsgroup.cpp` fix, so the Problems tab has no remaining sink). All three parts landed: the `Fl::event_inside(this)` guard, the `max <= 0` early-out, and the `dy == 0` horizontal-swipe case.

src/gui/constraintsgroup.cpp
tomburns: Applied in e48900d, together with `blocklistgroup.cpp` as suggested — so `ColorConstraintsEdit::handle` at `BlockList.cpp:811` is reachable again.

src/gui/view3dgroup.cpp
tomburns: Applied in e48900d — clamped against `slider->minimum()`/`maximum()` and `setSize` now takes the clamped `v`.

src/gui/configuration.cpp
tomburns: Resolved in bf19c0f — renamed the menu entry to `"Settings"` so it matches the dialog title, rather than reverting the title. Nothing else in `src/` or `doc_src/` refers to the old label.

tomburns: Went with the menu-side rename — pushed as bf19c0f, exactly the diff above.

tomburns:
All four findings from my review are now addressed on the branch:

| Finding | Status |
| --- | --- |
| `blocklistgroup.cpp` wheel sink (+ `max <= 0`, `dy == 0`) | e48900d |
| `constraintsgroup.cpp` wheel sink | e48900d |
| `view3dgroup.cpp` unclamped zoom accumulation | e48900d |
| `"Config"` menu vs `"Settings"` dialog title | bf19c0f |

For the last one I renamed the menu entry to `"Settings"` rather than reverting the dialog title, since the new title is the better name and nothing else in `src/` or `doc_src/` refers to the old label. That was the one change outside the original diff.

@akoehn — I have not built or run this locally (no meson/ninja on this machine), so CI is the only build verification so far. Happy to revert the menu rename if you would rather keep `"Config"`.

tomburns (approved):

akoehn:
I have zero opinion on naming etc. I do lack time to click through the gui parts that were touched though, would be good to have a visual inspection before merging.

tomburns:
I have since rectified my build environment and gave it a quick click through. seems to be good to go 👍

akoehn:
Thanks for checking, then we can safely merge 🙃

PR #46, opened 2026-09-05
Burrtxt: add json output

akoehn:
ported over but I am not 100% happy with the code and the windows-only ifdefs (were they tested?)

Putting here for visibility.


tomburns:
rebased onto latest main

PR #47, opened 2026-09-05
Triage and fix all first-party build warnings

tomburns:
Triages every warning the build emits, then fixes all of them that are ours.

A clean `meson setup build && ninja -C build` on master emits **339 warning lines**. That number is misleading — they collapse to **16 unique first-party sites** plus 3 kinds of linker warning, inflated because `src/lib/*` and `src/halfedge/*` compile separately for each of the four targets, and because one linker problem repeats once per object file.

After this PR: **6 warning lines, and zero from our own code.**

## The one that was an actual defect

`meson.build` set `-mmacosx-version-min=11.0` via `add_project_arguments`, which applies only to the main project. Neither meson subprojects nor CMake subprojects inherit it, so:

```
libfltk.a   ->  LC_BUILD_VERSION minos 26.0
libCatch2.a ->  LC_BUILD_VERSION minos 26.0
burrtools   ->  LC_BUILD_VERSION minos 11.0
```

`burrtools` advertised macOS 11 compatibility while statically linking code compiled against macOS 26 availability. On a real macOS 11 machine that is a plausible crash-on-launch rather than a graceful version error — the stated intent of the comment in `meson.build` was not being achieved.

Fixed with two mechanisms, because one is not enough:
- `add_global_arguments` instead of `add_project_arguments` (covers meson subprojects — catch2, and zlib/libpng on Windows)
- `CMAKE_OSX_DEPLOYMENT_TARGET` passed into the FLTK CMake subproject (inherits neither)

All three artifacts now report `minos 11.0`. The `11.0` literal is a single `macos_deployment_target` variable rather than repeated.

## Latent bugs — two hidden virtuals

`-Woverloaded-virtual` at `src/lib/voxel_2.h` and `src/gui/mainwindow.h`. Both benign today (every call site was checked and none resolves differently), but a trap for the next edit. Fixed with `using` declarations — the same pattern `voxel_2.h` already applies to `getMeshInternal` ten lines below.

## Dead code

- `placementFinder_c::prob` — stored, never read
- `layouter_c::resize`'s `wi`/`hi` accumulators — fed only commented-out code
- `fillPolyhedronHoles`'s `pairs_fixed` — a counter with no reader

## Variable length arrays

Seven VLAs (non-standard C++, a clang extension). The `assembler_1_c::getAssembly` ones mattered: sized by `rows.size()`, which grows with puzzle size and had no upper bound, so a large enough puzzle was a silent stack overflow. Those and `voxelframe`'s become `std::vector`.

`movementAnalysator_c::checkmovement` is on the disassembler's hot path, so a per-call allocation would be the wrong shape — its `check[]` becomes a member buffer allocated beside `movement`. Two things were checked first: every index is written before it is read within a call (all bounded by `next_pn`), and analysators are never shared across threads — each solve thread builds its own `disassembler_0_c`, each of which owns its own analysator. Decisively, `movement` and `matrix` are already per-object scratch buffers written the same way, so this adds no sharing that was not already there.

## Vendored code

`gl2ps.c` warns at a spot its upstream author already marked with a `FIXME`. Rather than patching vendored source that the next import would undo, the two third-party C files (`tr.c`, `gl2ps.c`) move into their own static library built with `-Wno-implicit-const-int-float-conversion`, passed through `get_supported_arguments()` so it is a no-op on GCC and the mingw cross build. Our own code keeps full `-Wall`.

## Verification

- Clean from-scratch rebuild (required — CMake caches the deployment target, so an incremental reconfigure proves nothing)
- `meson test -C build` passes 1/1
- Deployment targets re-checked with `otool -l`
- Because `assembler_1_c::getAssembly` directly produces solver output, the VLA change was **diffed, not just smoke-tested**: captured `burrTxt` output, reverted the three solver files, rebuilt, captured again. `DemoMirrorParadox` (assembler 1) and `PelikanBurr` (disassembler) are both **byte-identical**.
- No behaviour change is intended anywhere in this PR.

## What remains, and one thing worth knowing

The 6 remaining warnings are all from the linker, none from our source. Five are cosmetic FLTK link-flag noise. The sixth is not:

```
ld: warning: building for macOS-11.0, but linking with dylib
    'libpng16.16.dylib' which was built for newer version 26.0
```

That is Homebrew's **prebuilt** libpng — we do not compile it, so the deployment target cannot be applied. **The macOS 11 target is now honoured for everything we build, but still not end-to-end**; a genuine macOS 11 build would need a vendored or SDK-matched libpng. That is a packaging decision rather than a warning fix, so it is documented and left alone.

Also worth noting: this covers `-Wall` only (meson's `warning_level=1`, the weakest non-zero setting). Raising to `warning_level=2` would likely surface a batch of signed/unsigned comparison warnings — a separate decision, better made now that the output is readable.

https://claude.ai/code/session_015rcCvprH9hhQTFUypSSbf7


akoehn:
@tomburns I added -werror on github actions and fixed warnings that still popped up on my system (the joy of multiple compilers for a single language)

akoehn (dismissed):

akoehn:
@tomburns feel free to merge (I prefer rebase or squash merges but will let you decide)

akoehn:
adding -werror made the gh action for macos fail because libpng was dynamically linked from homebrew in the runner. Fixed that (static link instead), same problem class as Tom reported as far as I understand.

akoehn (approved):

PR #50, opened 2026-09-08
DO NOT MERGE rebased and cleaned up code from colin

akoehn:
this is here to create artifacts so others can test and give feedback.


grssam:
A few observations for the Rotation feature from this branch/build:

The default rotation solver does not account for the slight translations of the rotational axis that is required for the rotation to be possible in case there are neighboring voxels right next to rotational axis voxel. It rotates the piece assuming they have huge bevel/edge rounding. You can see that in the images attached that physically these rotations won't be possible.

I tested both the Andrew's rotational solver and the default rotational solver for my TIC clutch 4 and here are my observations:

- Andrew's solver returned back with 0 solutions. That's not quite right because previously I've ran this design against Andrew's original code and it does give a solution (let's call it A)
- Default solver did give a unique solution but it doesn't match A. The images below are from mid solve step where the pieces are digging in. When I tried to simulate the solution given by BT, the rotational steps where pieces were digging into each other did not rotate exactly as per the BT due to physical block. In this particular case of Clutch 4, it was still possible to do the rotational move using some mid rotation shimmy or slight translation due to the presence of empty space in the puzzle but this may not work for any dense TIC.

Both the solvers are not reliable at the moment. But they are still useful in many cases where the pieces might not be digging into each other during the rotational step. If BT can visually highlight this phenomena, then at least the user can be instructed to validate the solution on a physical copy. I am unsure of what's the percentage of false positives in cases where the rotational step can't be physically done at all but BT reports it possible..

image 
image

grssam:
Ported over (not as is) the 3D ViewCube and the transformation preview features in #99

PR #55, opened 2026-09-13
Add test coverage measurement

tomburns:
## What

Adds coverage measurement to BurrTools. There is currently none — no `gcovr`,
`lcov`, or `llvm-cov` in the build, the justfile, or CI — so no change can
demonstrate that it improved test coverage.

- `just coverage` — terminal summary
- `just coverage-html` — HTML report at `coverage-html/index.html`
- A non-blocking CI job that prints the summary and uploads the HTML as an artifact

Both are backed by `gcovr` against a separate `build-cov` directory configured
with Meson's builtin `b_coverage` option, filtered to `src/lib`, `src/tools`, and
`src/halfedge` so vendored code in `subprojects/` and `src/lua/` stays out of the
numbers.

**No production code changes.** `git diff master...HEAD -- src/` is empty.

## Baseline

**47.2% of lines (7618 of 16152)**, 50.7% of functions, 29.2% of branches.

Measured on macOS / Apple Clang / gcovr 8.6. Coverage numbers are
toolchain-specific — the line denominator itself varies between compilers — so
the Linux CI figure will differ. The CI number is the canonical one for comparing
across PRs, and this is documented in AGENTS.md.

## Two changes worth calling out

**A portability fix to `check-cppcheck`.** It used `-j$(nproc)`, and `nproc` is
GNU coreutils — it does not exist on macOS, so `$(nproc)` expanded to empty,
cppcheck got a bare `-j`, and `just check` was broken on any Mac. Since AGENTS.md
rule 6 requires `just check` to pass, this PR could not verify itself without
fixing it. Now uses POSIX `$(getconf _NPROCESSORS_ONLN)`, which works on both
platforms. Isolated in its own commit (`cc8d990`) if you would rather it went
separately.

**A design document** at `design/2026-09-13-test-coverage-stack.md`, describing the
four follow-up PRs this one enables.

## Notes

- macOS needs a `xcrun llvm-cov gcov` shim, since Apple Clang emits coverage data
  plain `gcov` cannot parse. The recipe applies it automatically via `os()`.
- The CI job is `continue-on-error: true` — it reports, it does not gate — and is
  skipped on tag pushes, where an instrumented rebuild would be pure waste.
- First `just coverage` run compiles all subprojects under instrumentation and
  takes several minutes. Subsequent runs are ~15s.
- Requires `gcovr` (`brew install gcovr` / `apt-get install gcovr`).

## Stack

PR 0 of five. The follow-ups add unit tests for voxel/symmetry, serialization,
disassembly internals, and the puzzle/problem model — each reporting its gain
against this baseline.


akoehn (commented):
I think it would be better to report coverage in a comment than in an html artifact that needs to be downloaded.

tomburns:
Done in 66aeb81 — the figures now go to the run summary and to a comment on the PR, and the comment is edited in place on later pushes rather than accumulating one per run.

One constraint worth flagging, since it shapes the implementation: a PR from a fork is given a read-only `GITHUB_TOKEN`, so the comment step cannot run there. This repo has open PRs from forks, so I guarded the step with a same-repo check rather than let it fail for exactly those contributors. Fork runs still get the identical report in the run summary, which needs no permissions and no download.

I kept the HTML artifact. gcovr's summary is only the three headline numbers, and the per-file drill-down is what you actually want once the headline tells you where to look. Happy to drop it if you'd rather not carry the upload.

Also worth recording now that CI has run: the canonical Linux/gcc number is **43.9% of lines (5721/13027)**, against the 47.2% (7618/16152) I measured locally on macOS/Apple Clang. The denominators differ too, not just the ratio — so the PR body's note about numbers being toolchain-specific turned out to matter more than I expected. The comment now says explicitly that the Linux figure is the one to compare across PRs.

akoehn (approved):

akoehn:
Squash?

tomburns:
Agreed — 15 commits for what amounts to a ~64-line functional change, and several of them add files that later commits delete. That churn shouldn't be in permanent history.

I'll squash it to two:

1. **`build: add test coverage measurement`** — the recipes, the CI job, the AGENTS.md entry, the design doc, the gitignore entry.
2. **`build: fix cppcheck job-count flag for macOS portability`** — kept separate deliberately. It touches a pre-existing recipe rather than anything to do with coverage (`-j$(nproc)` → `-j$(getconf _NPROCESSORS_ONLN)`; `nproc` is GNU coreutils and doesn't exist on macOS, so `just check` was broken there). Isolating it means you can drop or cherry-pick it independently if you'd rather it went separately.

If you'd sooner just use **Squash and merge** and not have me rewrite the branch, say so and I'll leave it alone — the result is the same and it saves a force-push.

One timing note: there's a stacked PR (#56) on top of this branch, so squashing here forces a rebase there. I'll do both together once #56's review settles, rather than rewriting the base out from under it mid-flight.

PR #56, opened 2026-09-13
Unit tests for voxel spaces, symmetries, bitfield and grouping

tomburns:
## What

Adds unit-level tests for four units that previously had none of their own: voxel
spaces, transformations and symmetry groups, `bitfield_c`, and `grouping_c`. Until
now these were exercised only indirectly, through full puzzle solves — so a defect
in voxel transformation or hash-table growth surfaced as a wrong move count in a
Pelikan burr regression, if it surfaced at all.

Also deletes `src/lib/bitfield_test.cpp`, `src/lib/voxel_0_test.cpp` and
`src/lib/main_test.cpp`. These were written against Boost.Test and were never added
to `meson.build`, so they have not compiled or run for as long as the Meson build has
existed — dead code that looked like test coverage. Their assertions are ported to
Catch2 first; nothing is lost.

**No production code changes.** `git diff coverage/tooling...HEAD -- src/` is exactly
the three deletions.

## Coverage

| | Before | After |
|---|---|---|
| Lines | 47.2% | **51.4%** |
| Functions | 50.7% | 53.7% |
| Branches | 29.2% | 32.8% |
| `src/lib/voxel.cpp` | 32% | **55%** |
| `src/lib/grouping.cpp` | 65% | **96%** |
| `src/lib/symmetries_2.cpp` | **0%** | **51%** |

Suite: 5040 assertions / 22 cases → **15119 assertions / 74 cases**, at a cost of
about 0.05s (~0.5% of suite runtime).

Measured on macOS / Apple Clang / gcovr 8.6, the same toolchain as the baseline in
the parent PR. The Linux CI figure will differ; per AGENTS.md it is the canonical one.

## Two bugs found, neither fixed here

Both are reported rather than fixed, so they can be reviewed on their own merits.

**1. `src/lib/symmetries_2.cpp:80` — sphere `transAdd` throws instead of returning `TND`.**
`symmetries.h:129` documents that `transAdd` may return `TND` when a composition does
not exist. `TND` is `(unsigned char)(-1)` = 255, and the sphere implementation carries
an extra `bt_assert(transMult[t1][t2] < NUM_TRANSFORMATIONS_MIRROR)` that the cube and
prism implementations omit — so it asserts precisely when the documented return is
required. 6912 of 14400 plain pairs are affected. Debug builds only; under `NDEBUG`
the assert is a no-op and the correct value is returned.

Suggestively, parsing the tables shows `tabs_0` and `tabs_1` contain *zero* `TND`
entries, while `tabs_2` contains 6912 — the only grid whose table encodes `TND` is the
only one that forbids returning it.

`test_symmetries.cpp` pins this with an exact counted skip, so when the assert is
removed the count drops to zero and the test fails, telling the fixer to delete the
workaround.

**2. `src/lib/grouping.cpp:136` — infinite loop.**
```c
for (unsigned int p2 = 0; p2 < p; p++)
```
declares `p2`, tests `p2 < p`, but increments `p` — the outer loop's variable. `p2`
never advances, so once `p > 0` the condition never becomes false. Reachable through
`disassembler_a.cpp:222-226`, which routinely puts several pieces in one set: a puzzle
using part groups can hang the disassembler with no error message. It has never fired
because every path to date arrives with `p == 0`.

`test_grouping.cpp` carries a file-level `KNOWN BUG` block describing the mechanism,
the reachability, and the `p == 0` invariant every case in that file preserves — so
the next contributor to extend it does not hang CI wondering why.

## Notes for review

- Tests are hybrid per the design doc: algebraic properties for the transformation
  core, worked examples for boundaries and error paths. No test asserts a solver
  iteration count (AGENTS.md rule 4).
- `test/test_helpers.h` is shared by the later PRs in this stack. Its ASCII-art axis
  mapping is proven against all five non-identity axis permutations, by positional
  assertions rather than bounds checks, so the guard survives `-DNDEBUG`.
- The sphere `transAdd` pin assumes `bt_assert` throws, which holds because Meson's
  `b_ndebug` defaults to false and CI does not set it. Under `-Db_ndebug=true` that
  case would fail confusingly. `test_solver.cpp:205` already had this dependency.
- Honest gap, documented in-file: sphere transformations 24-119 receive no composition
  coverage, because the fixture only survives `transform()` for the 0-23 cube subgroup.
- `identicalInBB` on prism and spheres cannot make `transform(0)` non-trivial —
  `voxel_1.cpp:47` and `voxel_2.cpp:33` both hardcode `if (nr == 0) return true;`.
- The spec also listed a puzzle deep-compare helper; it is deferred to PR 2, where the
  serialization tests actually need it.

## Stack

PR 1 of five, on top of the coverage tooling PR.


AGENTS.md
```diff
@@ -57,7 +57,7 @@ just build-tsan     # ThreadSanitizer (critical for solver data races)
 - **`src/gui/`**: FLTK-based graphical user interface and OpenGL 3D viewports (`mainwindow.cpp`, `view3dgroup.cpp`, `arcball.cpp`, `viewcube.cpp`).
 - **`src/halfedge/`**: Half-edge data structure for 3D polyhedron mesh manipulation and STL export.
 - **`src/tools/`**: XML parser/writer (`xml.cpp`), file existence helpers, and gzip stream wrappers (`gzstream.cpp`).
-- **`test/`**: Catch2 v3 automated regression test suite (`test_solver.cpp`) testing real-world puzzles across voxel grids, solver engines, and malformed input rejection.
+- **`test/`**: Catch2 v3 automated regression test suite across nine files (`test_solver.cpp`, `test_bitfield.cpp`, `test_cubemesh.cpp`, `test_grouping.cpp`, `test_manifold_smoke.cpp`, `test_minkmesh.cpp`, `test_symmetries.cpp`, `test_voxel.cpp`, `test_voxel_connect.cpp`) testing real-world puzzles across voxel grids, solver engines, malformed input rejection, voxel geometry, grid symmetries, and piece grouping.
```
akoehn:
```suggestion
- **`test/`**: Catch2 v3 automated regression test suite
```

man, these LLMs like to put unneccesary detail into everything.

akoehn (dismissed):
One change proposal, otherwise LGTM.

and why do we even have a custom bitfield implementation? I put that onto my todo list.

tomburns:
@akoehn change applied if you don't mind re-approving ✅

akoehn (approved):

akoehn:
Annoying edge case in the invalidation logic. Why would I not agree on a pr merge after my suggestions are applied

PR #58, opened 2026-09-13
Unit tests for XML serialization, gzstream and the puzzle roundtrip

tomburns:
## What

Adds unit-level tests for the XML writer, the XML parser, the gzip stream
wrapper, and a full puzzle save→load→compare roundtrip. None of these had
direct tests; they were exercised only by saving and loading whole puzzles,
where a defect surfaces as a corrupt file rather than as a clear failure.

The writer and parser take streams, so those tests run entirely in memory.
Only the gzstream and roundtrip cases touch the filesystem, and they use a
temporary directory that is removed even when an assertion fails.

**No production code changes.** `git diff coverage/voxel-symmetry...HEAD -- src/`
is empty.

## Coverage

| | Before | After |
|---|---|---|
| Lines | 51.4% | **52.8%** |
| Functions | 58.3% | **58.8%** |
| Branches | 33.0% | **33.3%** |
| `src/tools/xml.cpp` | untested | **65%** |
| `src/tools/gzstream.cpp` | untested | **84%** |
| `src/lib/puzzle.cpp` | — | **59%** |

Suite: 15119 assertions / 74 cases → **15264 / 109**, costing about 13ms of a
10.2s run.

The headline figure is not the interesting number — it is diluted by the whole
tree. The per-file numbers are: `xml.cpp` and `gzstream.cpp` went from untested
to 65% and 84%, and those are the files where a defect reaches a user as a
corrupt puzzle file rather than as an error message.

Measured on macOS / Apple Clang / gcovr 8.6, the same toolchain as the parent
branch's baseline. The CI job posts the canonical Linux figure.

## Four bugs found, none fixed here

Each is pinned by a test that **fails once the defect is repaired**, so the
pin cannot outlive the bug.

**1. A puzzle comment can be silently lost on reload.**
`xmlParser_c::next()` merges ignorable events by taking the lowest event code.
An element whose text is a single entity reference therefore reports `END_TAG`
instead of `TEXT`, and its text is dropped. `puzzle.cpp:329-336` only assigns
the comment when the parser reports `TEXT`, so a comment made only of
XML-special characters (`&`, `<`, `>`, `"`, `'`) vanishes on reload with no
error — `require(END_TAG)` even succeeds.

Likely a one-word fix: upstream kXML2 tests `ENTITY_REF` where `xml.cpp:1087`
tests `CDSECT`, and the `// ignorable` comment survived the port verbatim while
the constant did not. Applying that change locally makes the roundtrip correct
with no regressions in the rest of the parser tests.

**2. A literal `<` is accepted inside an attribute value.**
`pushText()` (`xml.cpp:844`) scans to the closing quote without stopping at
`<`, which XML 1.0 forbids in an attribute value. `<a name="v><b/>" z="2"/>`
parses cleanly, swallowing an entire child element into the attribute. expat
rejects both that and `<a name="x<y"/>`.

**3. `openGzFile` reports success for a file that does not exist.**
`gzstreambase` sets `badbit` correctly, but `std::istream`'s constructor then
calls `init()` and resets the state through their shared virtual `std::ios`
base. Callers cannot distinguish "missing" from "present but empty". The
plain-file fallback beneath it is also unreachable — it tests a pointer that
`new` cannot return. The GUI is unaffected (`mainwindow.cpp:2035` checks
existence first), but `burrTxt.cpp:203` and `burrTxt2.cpp:115` do not, so a
mistyped filename aborts rather than reporting a missing file.

**4. The GUI's Copy button loses a shape's name.**
Both `voxel_c` copy constructors (`voxel.cpp:116`, `:141`) omit `name` while
copying `weight` and the hotspot. `cb_CopyShape()` (`mainwindow.cpp:238`) uses
that path and `BlockList.cpp:330` renders the name, so copying a named shape
produces one without a name.

## Notes for review

- `bttest::puzzlesRoundtripEqual` is named for what it does rather than for
  equality in general: it compares only what a roundtrip must preserve, and its
  documentation lists what it skips — shape name, weight and hotspot, and all
  problem state beyond the name and result id. Two cases assert it returns
  **false**, so it has standing evidence it can fail rather than only passing.
- One case documents behaviour that is **correct but surprising**: an unquoted
  `=` inside an attribute value is ordinary text, so a lost quote yields a
  still-valid document with a different meaning. An earlier revision of this
  branch called that a defect; it is not, and expat agrees with BurrTools.
- Two deviations from the design doc, both deliberate: the new malformed-input
  cases are in-memory rather than new files under `test/malformed/` (strictly
  better, and the existing fixtures still cover the file path), and `TempDir`
  uses `std::filesystem::temp_directory_path()` because Catch2 v3 has no
  managed temporary-directory facility.
- Also spotted while reading, not tested here: `xml.cpp:131` dereferences
  `tagStack.rbegin()` without checking emptiness, so `endTag()` with no open tag
  is undefined behaviour; and every parser exception message reads `END_TAGat
  position` — a missing space at `xml.cpp:229`.

## Stack

PR 2 of five, on top of the voxel/symmetry tests.


tomburns:
Rebased onto the updated `coverage/voxel-symmetry`, and **dropped my AGENTS.md commit** — it's been made obsolete by c7abcb3.

That commit had bumped the directory map from "nine files" to "eleven". But c7abcb3 replaces the enumerated list with just:

```
- **`test/`**: Catch2 v3 automated regression test suite
```

which is the better fix by some distance. The list went stale on literally every PR that added a test file — mine included, one commit after the previous PR had updated it. Removing it solves that permanently instead of paying the maintenance forever. So the right move here is to drop my change, not to reconcile it.

This PR now touches `meson.build`, `test/test_xml.cpp`, `test/test_roundtrip.cpp` and `test/test_helpers.h` only. `AGENTS.md` is untouched.

Verified after the rebase: 15264 assertions in 109 test cases, all passing, and `git diff coverage/voxel-symmetry...HEAD -- src/` is empty.

akoehn (approved):

PR #59, opened 2026-09-13
Fix the defects the new test coverage uncovered

tomburns:
## What

The three test-coverage PRs ahead of this one reported defects without fixing them, so each could be reviewed on its own merits. This is that review.

Eight fixes, **45 lines across 9 production files**, one commit each so any can be dropped. Every fix lands with the test that proves it — the tests in the parent PRs currently *pin the broken behaviour*, and each is flipped here in the same commit as its fix.

## Merge order — please read, there is a live hazard

`master` right now contains this, added by #56:

```c
REQUIRE(thrown == (t == gridType_c::GT_SPHERES ? 6912u : 0u));
```

That assertion exists **because** sphere `transAdd` throws. Landing the one-line `symmetries_2.cpp` deletion without its paired test edit turns master red immediately. The same coupling applies to the other fixes.

This PR is based on **#58**, not master — `test_xml.cpp` and `test_roundtrip.cpp` don't exist on master yet and five of these fixes edit them. After #58 merges, this needs `git rebase --onto origin/master <58-tip>` before it will show a sane diff.

## The eight

| # | Fix | Effect |
|---|---|---|
| 1 | `grouping.cpp` unbounded loop | wrong group accounting, or a hang |
| 2 | `xml.cpp` entity-only body | puzzle comment silently lost on reload |
| 3 | `xml.cpp` `<` in attribute value | corrupt file loaded as valid-but-wrong |
| 4 | `symmetries_2.cpp` assert on `TND` | debug-only throw |
| 5 | `gzstream.cpp` missing file | CLI aborted on a typo |
| 6 | `xml.cpp` `endTag` empty stack | undefined behaviour |
| 7 | `xml.cpp` message spacing | cosmetic |
| 8 | `voxel.cpp` copy loses name | **see below — this one is a judgement call** |

**#8 is not a bug fix, and I've put it last so you can drop it with one rebase.** Both copy constructors omit `name` — but that was deliberate. They once carried an explicit `name(0)`, removed in `af0ce6a` only because `std::string` default-constructs empty, and `voxel_c::copy` still says so: *"we don't copy the name intentionally because the name is supposed to be unique."* Nothing enforces that uniqueness — names aren't keys, nothing looks up by name — and the visible result is that the GUI's Copy button hands back a nameless shape. But it reverses a stated intent, and it leaves `voxel_c::copy` inconsistent with the constructors. Your call.

## Things that are now different, not just better

- **#3 will reject some files that used to load.** Only damaged ones — BurrTools' writer escapes `<` in attributes — but a hand-edited or truncated file that previously loaded with markup absorbed into an attribute now fails outright. That's the point of the fix, but it is a behaviour change.
- **#2 also rejects a class of document.** An element whose body is a lone entity reference, walked via `nextTag()`, used to pass silently and now raises. Those elements are containers in this format and the writer can't produce one, so the risk is low — but it isn't purely additive, and the example-file gate can't see it.
- **#5 makes one GUI path silent.** `tryToLoad` returns `false`, but `cb_Load` discards it. In the narrow window where a file vanishes between the `fileExists` check and the open, the user now gets nothing where a load-error dialog used to appear.
- **#5 fixes the read side only.** `ogzstream` has the identical construction-order problem on the write side, which leaves `if (ostr)` vacuous at three call sites and lets `__rescue.xmpuzzle` silently write nothing. Out of scope here; worth its own issue.
- **#5 is asymmetric.** A *missing* file is now handled; a *corrupt* one still aborts the CLI through an uncaught exception. Deliberate scope, not an oversight.
- **#4 removes a check rather than correcting it.** `bt_assert(... < LIMIT || ... == TND)` would have kept the out-of-range guard. Given the table is generated and holds only values ≤239 plus `TND`, the practical loss is nil.

## Verification

- `--werror` build matching CI: 801/801 targets, zero compiler warnings
- `just test`: 15266 assertions / 115 cases · `just check`: exit 0
- All 18 `examples/*.xmpuzzle` load
- `./build/burrTxt /missing.xmpuzzle` → `could not open input file "..."`, exit 2 (previously aborted)
- No test anywhere still describes a defect fixed here

Assertion count drifts down slightly across the branch while cases rise. That's expected: a test cataloguing broken behaviour needs more assertions than the contract replacing it.


akoehn (approved):
No strong opinions about `voxel.cpp copy loses name` so I am approving

PR #61, opened 2026-09-14
Unit tests for the disassembly engine's internals

tomburns:
## What

Adds `test/test_disassembly.cpp`, covering the disassembly engine's internals:
`disassemblerNode_c`, the two node hash tables, the `separation_c` tree,
`disasmToMoves_c`, and the movement caches.

These were reachable before only through `test_solver.cpp`, which runs a whole
solve and checks the answer. That gives them plenty of *execution* but almost no
*assertions*: a node hash that silently dropped an entry, or an interpolator
that blended two keyframes wrongly, would still let the solver arrive at a
correct final answer. The tests here pin the contracts directly.

**No production code changes.** `git diff fix/coverage-found-defects...HEAD -- src/`
is empty.

## Coverage

| | Before | After |
|---|---|---|
| Lines | 53.0% | **54.3%** |
| Functions | 58.9% | **61.6%** |
| Branches | 33.4% | **34.1%** |
| `src/lib/disasmtomoves.cpp` | **0%** | **66%** |
| `src/lib/disassembly.h` | 16% | **95%** |
| `src/lib/disassembly.cpp` | 39% | **52%** |
| `src/lib/disassemblerhashes.cpp` | 87% | **99%** |

Suite: 15266 assertions / 115 cases → **17529 / 151**, adding about 0.02s to a
10s run.

`disasmtomoves.cpp` had no coverage at all: it is used by the 3D view's
disassembly animation, not by the solver, so nothing in the test suite reached
it. That is the one file here that goes from untested to tested.

Two files in this area were *already* well covered by the solver tests before
this PR — `movementcache_1.cpp` at 100% and `movementanalysator.cpp` at 84% —
and the line percentages for them do not move. The movement-cache tests are
still worth having: they assert that a query's answer is stable across repeats
and independent of the order queries are asked in, which is the memoisation
contract the solver relies on and never checks.

The uncovered remainder of `disassembly.cpp` is `state_c::save()`,
`separation_c::save()` and the parser's error branches — writing a disassembly
back out to XML. That belongs with the serialization work, not here.

## Test design

Fixtures come from the bundled example puzzles rather than from hand-built
structures wherever a real one exists, and they are read from **saved
solutions** — no solve runs in this file. A saved tree is repo data, so it
cannot shift underneath the tests the way solver output could if the search
order ever changed.

`examples/CubeInCage.xmpuzzle`'s tree is a pure `getLeft()` spine, which leaves
the two-child case dead, so `examples/Prisgon.xmpuzzle` problem 1 supplies a
branching tree for the walk and for piece reconciliation.

One case does build a `separation_c` by hand, and the comment above it says why:
`movesText()` appends the removed child's moves only when that child's subtree
runs longer than a single move, and no tree in `examples/` has that shape. Of
the six saved disassembly trees that ship there, two branch at all, and in both
the removed child is a single one-move separation.

A few of the assertions look stranger than they are, and carry comments
explaining the surprise:

- `disasmToMoves_c::getX/Y/Z` do **not** start every piece at the origin at step
  0, despite `disassembly.h`'s comment describing state 0 as "all values 0" —
  they return the pieces' real assembled grid coordinates.
- Piece 2's position reverses partway through CubeInCage's disassembly. That
  looks like a monotonicity violation; a dedicated case shows the reversal is
  real and not an artefact of where the steps were sampled.

## Stack

This is PR 4 of 5 in the coverage stack, based on #59.

- #55 coverage measurement (merged)
- #56 voxel, symmetry, bitfield, grouping (merged)
- #58 XML, gzstream, puzzle roundtrip
- #59 fixes for the defects that coverage uncovered
- **this PR** — disassembly engine internals
- next: the puzzle/problem model API


akoehn (approved):
Nice!

PR #65, opened 2026-09-14
Split the test suite into fast and slow, and test the Windows build

tomburns:
## What

Splits the test suite into a fast everyday loop and a slow stress suite, and
makes the Windows job actually run tests.

## The runtime problem

One case — `Minkowski mesher: random shapes` — accounted for roughly **85% of
the suite's runtime**: about 8.5s out of 9.7s. Everything else together finishes
in well under two seconds.

It already carried a `[stress]` tag, so this reuses that rather than inventing a
second convention. `meson.build` now registers two suites:

| Recipe | Runs | Test execution |
|---|---|---|
| `just test` | fast suite: Catch2 minus stress, plus the Python wrapper tests | **~1.6s** |
| `just test-slow` | the stress cases | ~9s |
| `just test-all` | both — **this is what CI runs** | ~8.9s |

Those are test-execution times. The recipes build first, so compilation is
extra and can dominate — measured on macOS with a warm `build/`:

| scenario | `just test` | `just test-all` | saved |
|---|---|---|---|
| nothing changed | 1.6s | 8.9s | 7.3s (82%) |
| one test file edited | 3.4s | 10.0s | 6.7s (66%) |
| one library `.cpp` edited | 3.1s | 10.3s | 7.2s (70%) |
| core header edited (wide rebuild) | 13.6s | 21.0s | 7.4s (35%) |

The saving is a constant ~7.3s — it is simply the stress case's runtime. What
changes is the fraction of wall-clock that represents, and on a wide rebuild
compilation does dominate.

**This makes CI slower, not faster.** The stress case now runs on three
platforms instead of two. The win is local iteration, plus the Windows job
running tests at all.

Two details worth calling out, because both would otherwise have quietly reduced
what gets run:

- **The Python wrapper tests are in the fast suite.** They take 0.13s. Left out,
  `just test` would silently cover less than it used to.
- **`just test-all` names both suites** rather than running everything meson
  knows about, so the vendored libpng subproject's own test suite cannot gate
  this project's CI.

The slow suite gets a 600s timeout. Meson's default is 30s, which the stress
case does not need today, but it scales with the `MINK_STRESS` environment
variable and CI runners are slower than a development machine — a timeout there
would read as a test failure rather than as a busy machine.

## The Windows problem

`build-windows` ran **no tests at all**. It built, stripped and packaged a
Windows artifact that nothing had ever executed.

It cross-compiles with MinGW on a Linux runner, so the test binary can't run
there directly. But `cross-mingw64.txt` already sets `exe_wrapper = 'wine64'`
with `needs_exe_wrapper = true`, and the build links `-static -static-libgcc
-static-libstdc++`, so the executable is self-contained. Adding `wine64` to that
job's dependencies is enough for meson to run the suite through it.

`build-linux` and `build-macos` already ran tests before packaging, but they
called `just test` — which now means the fast suite only — so they call
`just test-all` instead.

After this, **both suites must pass on all three platforms before anything is
stripped or packaged**, and `release` already depends on all three build jobs.

## Verification

- `just test`, `just test-slow`, `just test-all` all green locally; timings above
  are measured.
- `just check` clean.
- The Wine step could not be verified locally (this was developed on macOS); it
  is verified by this PR's own CI run.

## Follow-up, not in this PR

Wine tests the cross-compiled artifact, which is a good deal better than nothing
but is not the same as testing on Windows. A native `windows-latest` job would
be, and is a larger change — toolchain setup and probably some fresh build
breakage to work through.


akoehn (commented):
Is the test really that fast if the code has to be compiled? I would assume testing time is dominated by compilation time, not running the tests afterwards.

tomburns:
Fair challenge — the 1.7s figure was measured with nothing to rebuild, and the
PR body should have said so. I measured the cases that actually come up
(macOS, warm `build/`, times are the whole `just` invocation including the
ninja no-op check):

| scenario | `just test` | `just test-all` | saved |
|---|---|---|---|
| nothing changed | 1.6s | 8.9s | 7.3s (82%) |
| one test file edited | 3.4s | 10.0s | 6.7s (66%) |
| one library `.cpp` edited | 3.1s | 10.3s | 7.2s (70%) |
| core header edited (wide rebuild) | 13.6s | 21.0s | 7.4s (35%) |

So you are right for the wide-rebuild case: touch `voxel.h` and compilation
dominates, with the split saving 35% rather than 80%. You are not right for the
cases in between, which are the ones that come up most: re-running the suite
without touching anything is 5.5x faster, and editing a single file is 3x
faster. Compiling one translation unit and relinking costs about 1.5s here,
which is well under the 7.3s the stress case costs.

The saving is a constant ~7.3s in every row, because it is simply that one
case's runtime. What changes across rows is what it is a fraction of.

I have updated AGENTS.md and the PR body to give test-execution times and say
that compilation is extra, rather than quoting the no-op number unqualified.

Worth noting separately: the same 7.3s is paid on every CI run of every job, and
there it is pure addition with no compilation to hide behind. This change makes
CI run it on three platforms instead of two, so CI gets slower, not faster —
the win is meant to be local iteration, plus the Windows job actually running
tests for the first time.

tomburns:
@akoehn In addition to claude's commentary here; the separation will become more obvious as we grow the e2e test count, which is coming soon as we dive into the less unit-testable areas of the codebase. over time i expect compilation time will be dwarfed by test execution time in the slow suite

akoehn (approved):

PR #66, opened 2026-09-14
design: plan the next phase of test coverage work

tomburns:
## What

A design document recording what is left to do on test coverage, so the next
piece of work starts from measurement rather than re-deriving it.

Documentation only — no code, no build changes.

## Why now

The first phase (#55, #56, #58, #59, #61, #63) is merged. It moved engine line
coverage from 51.4% to 56.4% and fixed nine defects the new tests uncovered. The
obvious next question is "what else is worth testing", and the honest answer
needed a measurement pass that would be wasteful to repeat.

## What it contains

- **Where the uncovered code actually is**, by area rather than by file — 6,969
  uncovered lines, unevenly distributed.
- **Four packages** (P1–P4), each with its target files and current coverage,
  the approach, expected yield, cost, and the specific ways it can go wrong.
  Ordered by defect density rather than by line count: error and write paths
  first, grid parametrization second, the mesh/STL subsystem third, the
  assembler last.
- **What is deliberately not worth chasing.** The generated symmetry tables in
  `src/lib/tabs_*/` are a fifth of all uncovered lines. They are machine-emitted
  chains of `if (v.transform(N) && ...)`, one block per candidate transformation;
  covering them means constructing shapes exhibiting every symmetry of every
  grid, which is combinatorially large and finds nothing. Together with the
  threading, print and legacy-loader code, about 1,900 lines are not worth
  pursuing — which puts a realistic ceiling near 75%, not 100%.
- **The conventions the first phase settled on**, several learned by getting
  them wrong: every assertion must be able to fail, no commas or asterisks in
  test names, `#ifndef NDEBUG` around assertion tests, read saved solutions
  rather than solving, and keep production fixes out of coverage changes.

## On the numbers

Every figure is from macOS with Apple Clang, and the document says so up front.
CI reported 54.0% of 12858 lines for the same commit where Apple Clang reported
56.4% of 16000 — line counting is compiler-dependent, so the two are not
comparable and the Linux figure is the canonical one. Which areas hold the
uncovered mass is the same on either toolchain; the absolute counts are not.

## Note

It records a recommendation, not a commitment — the ordering is argued rather
than asserted, and P2 is flagged as needing its own design document before it
starts.


tomburns:
Committing this just for awareness; to be deleted once the plan is delivered. If we want to keep it out entirely I can just drop this from the PR list.

PR #69, opened 2026-09-16
test: parametrize the voxel suite across grids (coverage phase 2, P1)

tomburns:
Second package of the next-phase coverage plan (#66). #68 (P4) has merged, so this now targets `master` directly and its diff is only its own three commits. #70 and #71 remain stacked behind it.

## Coverage

Measured on macOS with Apple Clang over a freshly deleted `build-cov`. The first column is where #68 left off, not the original baseline.

| | `master` (after #68) | after this |
|---|---|---|
| Lines | 59.8% (9564/16004) | **62.2%** (9958/16004) |
| Functions | 69.7% (861/1236) | **70.8%** (875/1236) |
| Branches | 37.7% (6884/18251) | **39.6%** (7231/18251) |

Cumulative from the plan's 56.4% baseline: **+5.8 points of lines, +4.3 of branches.**

Per-file, against the plan's targets:

| File | before | after |
|---|---|---|
| `voxel_3.cpp` (rhombic) | 34% | **59%** |
| `voxel_4.cpp` (tetra-octa) | 34% | **74%** |
| `voxel_1.cpp` (prism) | 52% | **76%** |
| `voxel_2.cpp` (spheres) | 56% | **67%** |
| `voxel_0.cpp` (cube) | 70% | **75%** |

## The plan's estimate was 700–1,200 lines. This delivers ~394, and the reason matters

The plan modelled P1 as "mechanical enough to go quickly" — convert ~42 hardcoded `GT_BRICKS` sites to `ALL_GRIDS`. It also predicted the hazard, and the hazard turned out to dominate:

> A shape that is legal on the cube grid may be illegal or differently shaped on the rhombic or tetra-octa grids.

It is stronger than that. **An ASCII fixture is not a shape at all on the sparse grids.** `fromLayers()` sets a cell at every coordinate the art names, but `GT_SPHERES`, `GT_RHOMBIC` and `GT_TETRA_OCTA` accept only a subset — a 3×3×3 corner of the rhombic grid holds **six** valid positions, and `(0,0,0)` is not one of them.

Cells written outside that subset are not a differently-shaped piece; they are not a piece. Storage, indexing, bounding boxes and translation never consult `validCoordinate`, so those cases parametrize cleanly. Geometry does — which is why the mirror case does not merely fail on another grid, it **aborts** on `voxel_2_c`'s own `((hx+hy+hz) & 1) == 0`.

So roughly half the sites cannot be parametrized, and the fix is not a looser assertion. Run over `ALL_GRIDS`, "two separated voxels are not connected" fails *because on some grids they are*.

## What was done instead

**Parametrized (12 cases)** — the storage-level ones, where the property genuinely is shared.

**Given per-grid expectations (1 case)** — `minimizePiece`. The rhombic grid restores the corner's position mod 5 and rounds size up to a multiple of 5; tetra-octa uses mod 6 and multiples of 3. A 2×1×1 crop becomes 5×5×5 and 3×3×3. The case now separates what is uniform (the bounding box ends up exactly around the shape, no cell gained or lost, minimizing again is a fixpoint) from the per-grid size, with a table. A single expectation loose enough for all five would be satisfied by a `minimizePiece` that did nothing.

**Kept single-grid, with the reason recorded (9 cases)** — scaling, mirroring, connectivity.

**New cross-grid coverage that is actually true (11 cases)** — this is where the yield came from. Two new groups:

*Capability cases*, stating what differs: which coordinates hold a cell on each grid, that the origin is one on the dense grids and not the rhombic or tetra-octa ones, which grids implement `scale()`, which implement `scaleDown()`.

*Invariant cases* (`test_grid_invariants.cpp`), built on fixtures made the other way round — ask the grid which coordinates it accepts, then fill those. That gives every grid a legal shape, which is what makes a cross-grid assertion mean something on all five. The assertions are obligations, exactly as the plan asks:

- transforming preserves the cell count;
- every realisable transformation has an inverse, **found by search** because the inverse's number is grid-specific and hardcoding it is what a cross-grid case must not do;
- the neighbour relation is symmetric, and no cell is its own neighbour;
- `normalizeTransformation`'s result produces the same shape as the transformation asked for — asserted as that identity, so a stub returning 0 fails;
- `selfSymmetries` agrees with what `transform()` actually does.

Every one asserts its own premise first, so none can pass vacuously over an empty scan.

## A finding in the grids themselves

`GT_TRIANGULAR_PRISM` overrides `scale()` but **not** `scaleDown()` — it can be scaled up and never back down. `GT_SPHERES` overrides neither, exactly as `voxel.h` says ("spheres can't be scaled"). That asymmetry is invisible from the header's grid list and is precisely what a shared assertion would have hidden.

## Assertion count

The invariant scans cover a few thousand neighbour pairs per grid. The first draft asserted per pair and ran **24,782 assertions**; it now scans for the first violation and asserts once, for 1,533. Same coverage, and the failure still names the offending cell through the `INFO`. The plan's "one fact, one assertion" convention is the reason.

## Mutation-checked

Stubbing `normalizeTransformation` to 0; making the cube grid's face neighbours asymmetric; making a cell its own neighbour; making `validCoordinate` reject everything; making the base `scaleDown` claim success; making the base `scale` resize; making the rhombic `validCoordinate` accept everything. Each turns the relevant case red.

`just test-all` passes. No production code is changed.


akoehn:
Testing some different models, this is what it came up with :)

P2 — Scale-down capability expectations misclassify two grids and never validate them

Head lines: test/test_voxel.cpp:799–805, 819–831  

GT_RHOMBIC and GT_TETRA_OCTA are marked canEverScaleDown = true, but their overrides unconditionally return false (src/lib/voxel_3.cpp:449–452, src/lib/voxel_4.cpp:302–305). The test passes because it only checks the expected answer for entries marked false; otherwise it compares two calls against each other.  

Concrete missed regression: making either unsupported override return true without changing anything still passes this case, falsely advertising successful scaling. Mark both grids unsupported and assert the actual answer against the table for every grid, using a known-scalable fixture for the supported case.


P2 — The single-cell connectivity test exercises an empty shape on two grids  

Head lines: test/test_voxel_connect.cpp:205–209  

(1,1,1) is invalid for spheres (odd coordinate sum) and rhombic grids (fails their coordinate predicate). countState() counts storage slots regardless of validity, whereas both unionFind() and connected() filter through validCoordinate() (src/lib/voxel.cpp:425–428, 485–487). Thus the count assertion passes, but connectivity finds zero participating cells and returns the empty-case result.  

Concrete trigger: the existing GT_SPHERES and GT_RHOMBIC iterations already bypass the singleton path entirely, so they cannot detect the singleton-handling failures this case promises to catch. Choose a valid coordinate per grid and assert its validity before filling it. The parametrized block at test/test_voxel_connect.cpp:17–23 has the same empty-fixture problem for rhombic and tetra-octa grids.

tomburns:
Both confirmed and fixed in `d137e02b`. The model earned its keep — these were real, and the second one generalised further than stated.

**Scale-down capability expectations.** Correct on every point. `voxel_3_c::scaleDown` (voxel_3.cpp:449) and `voxel_4_c::scaleDown` (voxel_4.cpp:302) are stubs returning false unconditionally; only `voxel_0_c` implements it. The table's premise was that overriding the method implied implementing it, and the assertion structure — checking the expected answer only on the `false` rows, otherwise comparing the call against itself — meant nothing ever tested that premise.

Verified your exact scenario: making `voxel_3_c::scaleDown` return `true` passed the old case. It now fails with `true == false`. The two rows are corrected, the answer is asserted against the table for every grid, and the supported grid must now actually *perform* the scaling rather than only advertise it.

There was a third defect underneath that one you didn't call out: the fixture was empty on GT_RHOMBIC too. `makeVoxel(2,2,2) + setAll + scale(2)` leaves `countState(VX_FILLED) == 0` there, because `scale()` drops cells the grid doesn't accept. So the old case was asking rhombic whether it could scale down nothing.

**Single-cell connectivity.** Confirmed, with the mechanism exactly as you describe. Probed the validity directly:

| grid | `(1,1,1)` valid |
|---|---|
| GT_BRICKS | yes |
| GT_TRIANGULAR_PRISM | yes |
| GT_SPHERES | **no** |
| GT_RHOMBIC | **no** |
| GT_TETRA_OCTA | yes |

`countState()` counts storage slots, `connected()`/`unionFind()` filter through `validCoordinate()` (voxel.cpp:425, voxel.cpp:485), so the count assertion passed and connectivity then took the empty-shape branch.

Your note about the parametrized block at test_voxel_connect.cpp:17-23 was right as well, and worth quantifying — the `{"##","##"}` fixture yields `countState == 4` on every grid but the number of cells the grid *accepts* is 4, 4, 2, **0**, **0**. So "a solid block is face connected" was asserting the vacuous empty-shape answer on two of the five grids — the very thing the case below it exists to cover separately.

**What changed.** Rather than adding a coordinate table, I promoted `validCoordinates()` and `legalShape()` out of `test_grid_invariants.cpp` into `test_helpers.h`, alongside a new `countValidFilled()`. That file already encoded precisely this rule ("ask the grid which coordinates it accepts, and fill those") and documented the hazard in its header comment; the connectivity and scaling cases simply weren't using it. A second copy would have been a second thing to get wrong.

Every affected fixture now asserts it is non-vacuous before the property is asserted over it. Full suite green: 278 cases, 22255 assertions.

akoehn (approved):

PR #70, opened 2026-09-16
test: cover the mesh and STL subsystem (coverage phase 2, P2)

tomburns:
Third package of the next-phase coverage plan (#66). **Stacked on #69**, which now targets `master` since #68 merged — review #69 first.

The parent plan calls P2 "the one package that deserves its own design document rather than being started directly from this one." The first commit is that document; the other five implement it in the order it sets out.

## Coverage

| | after #69 | after this |
|---|---|---|
| Lines | 62.2% (9958/16004) | **69.5%** (11129/16011) |
| Functions | 70.8% (875/1236) | **80.0%** (994/1242) |
| Branches | 39.6% (7231/18251) | **43.4%** (7922/18255) |

Cumulative from the plan's 56.4% baseline: **+13.1 points of lines, +14.1 of functions, +8.1 of branches.** This is the largest single package, as the plan predicted.

## Why the design document

P4 and P1 were retrofits: the code had a testable API and the job was to call it. Here the difficulty is not reaching the code but knowing what to assert once you have. A mesh is a large, floating-point, order-sensitive artefact, and the tempting test is to record one and compare — which the plan is explicit about:

> a recorded-output test will either be brittle or will be "fixed" by re-recording, which silently ratifies whatever regression prompted it.

So nothing here records a mesh. Every assertion is an identity, an invariant, or a roundtrip.

The document also records the package's central trap: `Polyhedron::check()` already exists and walks every invariant, but it is `bt_assert`-based, so it compiles to nothing under `NDEBUG`. A case whose only assertion is `check()` silently tests nothing in a release build. Every use of it here is `#ifndef NDEBUG`-guarded and paired with assertions that hold in both build types.

## What was found

**`Vector3D::operator*(const Vector3D<T>* mat)` is broken.** Its inner loop is `V[i] += mat[i][j]` — it never multiplies by anything, so it returns the matrix's row sums and discards the vector entirely. Every vector times a given matrix comes back identical. Nothing calls it, nor its companion `transpose()`, so the defect is latent. Not fixed: without a caller there is nothing to settle whether the intended convention is row- or column-major, which is what decides the correction.

**`stl.cpp`'s own `basename()` calls `strchr` where it needs `strrchr`.** It is used on Windows, macOS and Emscripten, and returns everything after the *first* separator — so every STL a macOS or Windows user exports carries most of their directory structure in its header, and binary STL's 80-byte header truncates a deep path part-way. CI runs on Linux, where the POSIX `basename` is correct, which is why it has gone unnoticed. The case splits by platform so it passes on both.

**`mergeCoplanarFaces` returns a mesh that is not twin-linked.** It builds a fresh polyhedron and never calls `finalize()`; its header documents the twin-link requirement only on the *source*. Its one caller walks face edge loops and never touches a twin, so nothing has noticed — but adjacency, `numHoles()` and `check()` all need twins, and the signature does not say so.

**The coplanar merge in `stlExporter_c::write` does nothing for these meshers.** Measured over a single cell, an L-tromino and a plus pentomino on two grids: the merged mesh has *identically* as many faces and triangles as the mesher produced, in all six combinations. The function is not broken — `test_modifiers.cpp` shows it collapsing a four-triangle fan to two — these meshers just emit groups already at their minimal triangulation. So it costs a full polyhedron rebuild per export and currently buys nothing. A case asserts the equality so that if a mesher ever starts producing mergeable geometry, someone finds out.

**`distanceToPlane`'s sign is the reverse of the usual convention** — `(P - *this) * N` measures from the point towards the plane, so a point on the normal's side gets a negative distance.

All five are pinned where a reader will meet them and raised separately, per the convention #59 and #64 set.

## Fixtures built for the job

`mergeCoplanarFaces` needed a fixture designed against it: a plain cube is no good, because each face is already two coplanar triangles and merging a pair into a quad gives two back. The fixture is a cube whose top is a fan of four triangles around an added **interior** vertex — interior because the header promises to keep every boundary vertex, so a subdivision adding boundary vertices would not reduce either. A case asserts the fixture really is four coplanar faces before anything relies on it.

The sphere mesher reuses P1's finding: a sphere-grid shape must be built from coordinates the grid accepts, not from ASCII art.

The STL reader is written from the format, not from the writer — the design doc is specific that doing it the other way lets the two agree on a shared misunderstanding. It gets its own cases, including one proving a truncated document raises rather than yielding fewer triangles.

## Assertion counts

Two drafts ran 24,782 and 33,483 assertions by asserting per element. Both now scan for the first violation and assert once — 1,533 and 170 — for identical coverage, with the `INFO` still naming the offending element. The plan's "one fact, one assertion" convention.

## Mutation-checked

~20 deliberate breaks, each verified to turn the relevant case red: negating the cross product, removing `angle()`'s clamp before `acos`, swapping `setSpheric`'s angles, dropping `volume()`'s division by three, negating a face normal, making `valence` over-count, making the three-argument scale apply x to every axis, making the merge never accept a group, reversing the STL triangle winding, zeroing the binary triangle count.

`just test-all` passes. No production code is changed.


akoehn:
P2: Temporary directories collide across test runs (test/test_stl.cpp:40–45, test/test_spheremesh.cpp:44–48). Names use process-local counters in the shared temp directory, and construction deletes any existing directory. Concurrent runs can delete each other’s exports. Use exclusively created, process-unique directories.

P2: Idempotence test skips the second merge (test/test_modifiers.cpp:330–331). once has no twin links, so the second call treats every face as an isolated group and merely copies it. Call once->finalize() before merging again to exercise actual idempotence.

[I have no idea what this means but probably opus will figure it out]

tomburns:
Both confirmed and fixed in `d473e932`. (On the second one — it means the idempotence test was never running a second merge. Detail below.)

**P2 — temp directory collisions.** Correct. `TempDir` named itself `bttest_stl_<process-local counter>` in the shared system temp directory and called `remove_all()` on that path *before* creating it. Two test binaries running at once both reach `bttest_stl_1`, and the second one's constructor deletes the first one's exports mid-run, while the first is still writing to them.

Fixed as suggested: a random token drawn once per process is mixed into the name, and `create_directory`'s "did I create it, or was it already there?" answer drives a retry instead of destroying whatever it finds. Four concurrent suites now run clean and leave nothing behind.

One wrinkle worth recording, because it bit immediately. My first attempt used a longer token and broke the binary-header case. The binary STL header is a fixed 80-byte field, and on this branch — before #74's basename fix — the exporter writes the *whole path* into it. The old name was already only about seven characters clear of the limit on a normal macOS temp path:

```
/var/folders/wq/3pyknw4j31g0z4rfvdklmltm0000gn/T  →  48 chars
.../bttest_stl_1/counted.stl                      →  73 of 80
```

So the new name is deliberately short — a three-character tag plus eight hex digits, the same length as the `bttest_stl_` it replaces — and the header case now asserts the path fits and says why, rather than failing as an unexplained string mismatch. That sensitivity disappears on #74, where the exporter writes a basename; the rebase resolved in favour of the plain `m.name == "counted.stl"` there.

The two `TempDir` copies stay duplicated rather than being promoted to `test_helpers.h`. Both are fixed identically. I left it alone because that header is modified by two later branches in this stack and the merge cost outweighed the tidiness — happy to consolidate it if you'd rather.

**P2 — idempotence skipped the second merge.** Correct, and this one is a genuinely good catch. `mergeCoplanarFaces` groups coplanar neighbours by walking half-edge `twin()` links (modifiers.cpp:1024-1026), and builds its result with `addFace`/`copyFace` without joining the twins afterwards. So `once` comes back with **no twin links at all**. Fed straight back in, every face looks like an isolated group of one, the `groups[g].size() > 1` branch is never taken, and the second call just copies each face across. Equal face counts and volumes then say nothing about idempotence — they are exactly what a copy produces.

`finalize()` is what joins the half-edges, so calling it first is what actually asks the merge to run twice. Done.

For the record, the merge does turn out to be genuinely idempotent once the second pass really runs, so this strengthens the case without changing any behaviour — which is the outcome you'd want, but was not knowable before.

363 cases, 23,341 assertions, green. The three branches above this one are rebased and pushed.

akoehn (approved):

PR #71, opened 2026-09-16
test: cover the assembler's validation surface (coverage phase 2, P3)

tomburns:
Fourth and final package of the next-phase coverage plan (#66). **Stacked on #70 → #69**, which targets `master` since #68 merged — review those first.

## Coverage

| | after #70 | after this |
|---|---|---|
| Lines | 69.5% (11129/16011) | **71.5%** (11441/16011) |
| Functions | 80.0% (994/1242) | **81.0%** (1006/1242) |
| Branches | 43.4% (7922/18255) | **44.9%** (8199/18255) |

Per-file: `assembler_0.cpp` 48% → **51%**, `assembler_1.cpp` 57% → **59%**, `assembler.cpp` → **88%**.

**Cumulative across all four packages: 56.4% → 71.5% lines, 65.9% → 81.0% functions, 35.3% → 44.9% branches.**

The plan's target was "near 75%" for covering 60% of what it judged worth chasing. This lands at 71.5% with P4, P1, P2 and P3 all delivered.

## Most of this needs no solving

The plan puts the assemblers last because covering them needs runtime. But `createMatrix` validates *before* any search begins — it counts cells, checks every piece has somewhere to go, and refuses puzzles the assembler cannot express — and all of that is reachable from a bar puzzle built inline. Those cases stay in the fast suite, which is unchanged at 1.7s. The cases that solve are tagged `[stress]`.

Covered without solving: too many units, too few units, variable cells absorbing a shortfall (paired with the rejection, so neither is consistent with a check that never fires), a piece that fits nowhere, `assembler_0`'s refusal of piece ranges against `assembler_1`'s acceptance of the same problem, `findAssembler`'s choice between them, and that all seven error states carry distinct non-empty messages.

## Four wrong expectations, each worth the correction

**`setShapeMaximum` alone creates a part with minimum 0** — the range `[0..max]`, not "exactly one". Every fixture built that way was refused by `assembler_0` as unhandable, so every case returned the wrong error. Worth knowing for anyone building a problem programmatically.

**A six-cell bar filled by a 2-bar and a 4-bar has two arrangements but ONE assembly.** The result is symmetric under the half-turn mapping x→5−x, which carries one arrangement onto the other, and the rotation reduction discards the duplicate. The early-exit cases needed three distinct bar lengths so there is genuinely more than one to find.

**`stopped()` means "not currently running", not "was asked to stop"** — it is `return !running;`, true before a search and true again after one. And **`stop()` before `assemble()` does nothing at all**: `assemble()` stores false into the abort flag on entry, silently dropping a request made beforehand. Both pinned, with a separate case showing `stop()` from inside the callback does halt the search.

**Solid Six Piece Burrs records 1063 assemblies; a fresh solve found 588.** Not a defect — the recorded count depends on which reductions were enabled when the file was saved, and this one kept mirror solutions. 1063 is exactly what `keepMirror` gives. That turned into a better test than the one it replaced.

## The bundled puzzles as oracles

Twelve problems across nine puzzles, two assemblers and three grids, re-solved and compared against counts that came from the repository rather than from this code — the plan's own recommendation, and the reason none of these expected values had to be invented.

A companion case pins the reduction ordering: relaxing a reduction can only ever report *more* assemblies, because each reduction is a merge. It takes two puzzles, since no single cheap one exercises both flags — Solid Six Piece Burrs for mirror (588 → 1063, the second being the recorded figure), Hex Sticks for rotations (33 → 744). This is the only place `createMatrix`'s `keepMirror` and `keepRotations` arguments are exercised anywhere in the suite.

## One regression I introduced and fixed

The first version of that case ran Solid Six Piece Burrs with `keepRotations`, finding 12,597 assemblies. Under gcov instrumentation the dancing-links inner loop in `assembler_1.cpp` then accumulates **over four billion hits**, and gcovr rejects a counter that large as a parse error (gcc bug 68080) — so **`just coverage` failed outright**.

Hex Sticks demonstrates the same property for a thousandth of the work, so the expensive run is gone and the reason is recorded in the case so nobody reintroduces it. `just coverage` completes cleanly, with no warnings.

## Mutation-checked

Removing the too-many-units check, making the too-few-units check ignore variable cells, removing the cannot-place check, making `canHandle` accept everything, ignoring the callback's return value, and ignoring the abort flag — each turns the relevant case red.

`just test-all` passes (fast 4.5s, slow 48s). No production code is changed.


akoehn:
P2 — Placement diagnostic assertion uses the wrong identifier 

test/test_assembler_config.cpp:182–184  

getErrorsParam() returns the puzzle’s shape ID, not its problem-local part index: assembler_0.cpp:686 returns the negated shape ID, and createMatrix() negates it back. The GUI likewise uses it to select a shape. This assertion passes accidentally because longBar has both identifiers equal to zero.  

Concrete scenario: insert an unused shape before the pieces. The unplaceable bar then has shape ID 1 but part index 0; the correct assembler response fails this assertion. Assert against longBar and make the two IDs differ in the fixture.


P2 — Hole-limit test never exercises the configured limit  

test/test_assembler_config.cpp:504–516  

The test sets maxHoles, then uses assembler_0_c, which never reads that setting. The actual configuration branch is in assembler_1.cpp:687–692, and is reached only when total minimum and maximum piece volumes differ. Consequently, the comment that this limit constrains this search is incorrect.  

Concrete scenario: removing assembler1’s configured-hole-limit branch leaves this test unaffected, even though ranged-piece puzzles would admit assemblies exceeding the requested limit. Exercise assembler1 with piece ranges and variable result cells, including a solution excluded by the limit.


P2 — Factory-selection assertions do not identify the returned assembler  

test/test_assembler_config.cpp:234–240, also :721–731  

Calling the static assembler_0_c::canHandle() checks the input, not the factory’s returned object. Both assemblers accept the initial fixed-count problem. The later coverage check similarly infers implementation selection from eligibility without inspecting a factory result.  

Concrete scenario: changing findAssembler() to always return assembler_1 makes both sections of the factory test pass, while the coverage check still claims both implementations were exercised. Check the returned implementation directly or use behavior that distinguishes it.

tomburns:
All three confirmed and fixed in `5b300843`. I reproduced each one as a mutation against the pre-fix tests before changing them, so the claims below are checked rather than argued.

**P2 — placement diagnostic identifier.** Correct. `prepare()` returns `-getShapeIdOfPart(pc)` (assembler_0.cpp:686), `createMatrix` negates it back (assembler_0.cpp:748), and mainwindow.cpp:3141 does `selectShape = getErrorParam()` — it's a shape id throughout.

Your suggested scenario, verbatim: inserting an unused shape ahead of the pieces made the old assertion fail with `1 == 0`, confirming it had been passing only because `longBar` was shape 0 and part 0 at once. The fixture now carries that unused shape permanently, the assertion is against `longBar`, and there's a second assertion that the two identifiers actually differ, so the fixture can't quietly drift back to the degenerate case.

**P2 — hole limit.** Correct, and stronger than stated: `maxHoles` appears *nowhere* in assembler_0.cpp, so assembler_0 doesn't merely fail to reach the branch, it has no notion of the setting at all. The live branch is assembler_1.cpp:689-690, and as you say it's reachable only when total min and max piece volumes differ — with `min == max` the count is pinned to `res_filled - min` and the configured limit is never consulted. The comment claiming the limit constrained that search was wrong.

Built the ranged case you asked for: a 1×1 piece with range [3..5] against a five-cell result of three filled and two variable (`###++`). That gives four assemblies with 2, 1, 1 and 0 holes, so the limits separate cleanly:

| maxHoles | assemblies |
|---|---|
| unset | 4 |
| 0 | 1 |
| 1 | 3 |
| 2 | 4 |

Disabling the branch now fails the case. I kept a reduced version of the old test as its own explicit claim — that assembler_0 ignores the limit entirely — because a passing "the limit changed nothing" reads like the limit was honoured and happened to exclude nothing, which is exactly the misreading that hid this.

**P2 — factory selection.** Correct. Your scenario understates it slightly: I deleted `findAssembler`'s assembler_0 branch *outright* rather than redirecting it, and both sections plus the coverage check still passed — 29 assertions, no failures.

Both sites now identify the returned object with a `dynamic_cast`. The original comment explicitly preferred "behaviour rather than a dynamic_cast", which is a reasonable instinct, but the behaviour it picked was a static `canHandle()` call describing the *problem*, so it held whatever the factory did with the answer. Neither assembler exposes a name or version to identify itself by, so the cast is the only thing available that inspects what came back. The bundled-counts check now asks the factory and inspects the result instead of inferring the implementation from eligibility.

Full suite green: 383 cases, 23602 assertions.

akoehn (approved):

PR #74, opened 2026-09-16
fix: the defects the phase-2 coverage work uncovered

tomburns:
Fifth in the coverage stack, and the counterpart to it. **Stacked on #71 → #70 → #69** — review those first.

The four coverage packages deliberately changed no production code: they pinned each defect where it was found and recorded why the fix belonged elsewhere. This is elsewhere. Each fix is its own commit, so any can be dropped alone.

Every fix **flips a test from pinning the broken behaviour to guarding the correct one** — the cases already existed, they just asserted the wrong thing.

## Fixed

**1. `stl.cpp`'s `basename()` returned nearly the whole path.** It called `strchr` — the *first* separator — where a base name needs `strrchr`. `"/tmp/a/b.stl"` yielded `"tmp/a/b.stl"`. That name goes into the exported file, so every STL written on Windows, macOS or Emscripten carried most of the user's directory structure in its header, and binary STL's 80-byte header silently truncated a deep path part-way. Linux uses the POSIX `basename` and was unaffected, which is why CI never saw it. Its `res1 > res2` comparison of pointers into different objects was also undefined behaviour. The test's platform split is now gone — the point of the fix is that all platforms agree.

**2. `Vector3D::operator*(const Vector3D<T>* mat)` discarded the vector.** `V[i] += mat[i][j]` never multiplied by anything, so it returned the matrix's row sums; every vector times a given matrix came back identical. #70 pinned rather than fixed it because with no caller there was nothing to settle row- vs column-major. Row-major is what makes it consistent with the `transpose()` directly below it, and a test now asserts that relationship — `v * transpose(m)` must equal the column-wise product — rather than either convention alone.

**3. `xmlParser_c::getPositionDescription()` was declared and never defined.** Removing a declaration from a public header is an API change, which is why #68 left it. It is safe here for a specific reason: nothing can be linking against a symbol that has no definition.

**4. `distanceToPlane` reported the opposite sign to the convention.** A point on the normal's side got a negative distance. This is the one judgement call — a convention rather than an arithmetic error — so it is **last and separately droppable**. Safe because nothing calls it, so no caller was compensating.

## Test fixed, not code

**`test_immediate_stop` was racy** and failed once on CI during this stack's own verification.

I diagnosed this wrong the first time and it is worth recording. It looked like the assembler clearing its abort flag on entry to `assemble()` — which is real, and documented in #71 — but that is not the cause. The binding already guards `stop_requested` at five points including inside the callback, and the solver's reset is *load-bearing for resume*: `solvethread.cpp` reuses a stopped assembler without calling `createMatrix` again, so removing it would break resuming a paused solve.

The actual cause is simpler: the worker starts solving at **construction**, and `stop()` does not discard what it has already queued. Measured — no delay before `stop()`: 300/300 raise `StopIteration`; a 1 ms delay: 99/100 return a result instead.

Returning already-found results is defensible, so the test changed. It now asserts what is actually guaranteed — the iterator terminates, stays exhausted, and does not run to completion. That last part is timing-independent because the queue is bounded at `MAX_QUEUE_SIZE = 4` and the worker blocks when full: with no consumer it cannot pass four results, and the fixture has twelve. Across delays from 0 to 500 ms the count is 0 or 4, never 12.

## Deliberately not fixed

Four findings stay documented rather than changed, because each is a design or performance decision rather than a defect:

- **`separation_c` vs `separationInfo_c` `getNumSequences()`** return different quantities (6 vs 13) for one disassembly. Changing either alters the meaning of stored data.
- **`mergeCoplanarFaces` returns a mesh that is not twin-linked.** Adding `finalize()` costs every STL export for a caller that does not need it; documenting the requirement may be the better fix.
- **The coplanar merge in `stlExporter_c::write` is a no-op** for the current meshers — a full polyhedron rebuild per export that changes nothing. Removing it is a performance call.
- **`defineEntityReplacementText`'s unreachable guard** is harmless; removing it is churn.

## Verification

`--buildtype=release --werror` (the CI configuration, not just `just test`): builds clean, 382 cases / 23,522 assertions pass, Python suite passes.


test/test_stl.cpp
```diff
@@ -297,53 +296,27 @@ TEST_CASE("stl export: the solid name is derived from the path -- and on this pl
...
+     the FIRST separator. That is fixed, so there is one expectation again.
+
+     No platform guard now: the point of the fix is that all of them agree. */
   REQUIRE(m.name == "named.stl");
-#endif
+
+  /* a path with no separator at all must come back unchanged rather than
+     losing its first character -- the branch where strrchr finds nothing */
+  {
+    const std::string bare = "bare.stl";
```
akoehn, line 393: unless I miss something, this is referencing a file in a non-temp directory, and then writing to it. Not a problem in most of the cases but still not good.

Might be a problem in other tests as well? Have not checked yet

tomburns: Confirmed and fixed in `ef3f994b`.

It wrote `bare.stl` relative to the process's working directory, which is the source root when the suite runs. The bare name isn't incidental — the case exists to cover the branch where the exporter's name handling finds no directory separator, so the path can't just be moved under the temp dir like every other write in the file. The process moves instead, via an RAII guard that restores the previous directory even when an assertion throws.

That also closed a smaller hole you didn't mention: the explicit `remove()` at the end of the case was skipped on failure, so a *failing* run left the file in the checkout.

On "might be a problem in other tests as well" — I swept the suite, and this was the only one. Every other write goes through `TempDir`. The one apparent exception, `test_minkmesh.cpp:94`, writes to a directory the `MINK_DUMP` environment variable names and is off by default.

The guard is deliberately narrow, since the working directory is process-global and the rest of the suite loads fixtures through relative paths (`examples/...`) that only resolve from the source root.

akoehn (approved):

PR #75, opened 2026-09-17
test: close the coverage gaps mutation testing found in the merged suite

tomburns:
Stacked on #74. Test-only.

An audit of everything merged to `master` in the last month found tests that pass without testing what they claim. This closes the ones that have an honest test-side fix, and corrects the ones that do not.

Findings are not opinions here. Each one is a **mutation applied to production code that the full suite failed to notice**, and each fix is verified by re-applying that mutation and watching the new test fail.

### Mutations that used to survive, and now do not

| Mutation to production code | Before | After |
|---|---|---|
| `voxel_c::getSTLMesh` → always the flat fallback | survived | killed |
| `puzzle_c::removeColor` → drop both loops | survived | killed |
| `problem_c::exchangeShapes` → drop the result renumber | survived | killed |
| const `problem_c::getAssembler` → `nullptr` | survived | killed |
| `separation_c::exchangeShape` → positional swap | survived | killed |
| `disassemblerNode_c` → `waylength = step` | survived | killed |
| `voxel_c::getMirrorTransform` → search proper rotations | survived | killed |
| `voxelTable_c` → delete the `PAR_MIRROR` filter | survived | killed |
| `voxel_2_c::validCoordinate` → `true` | survived | killed (6 cases) |
| `voxel_2_c::onGrid` → `true` | survived | killed |
| `voxel_3_c::onGrid` → `true` | survived | killed |
| `getAttributePrefix` → stride 8 | survived | killed |
| bitfield hex letters `a`–`e` shifted | survived | killed |
| `moRehash` → silently drop a bucket | survived | killed |

The worst of these is the first: `getSTLMesh` falls back to `getFlatMesh()` when the Minkowski mesher fails, and every view assertion — face tagging, normals, containment — is satisfied identically by that fallback. The chamfer mesher could be deleted from the 3D view outright and nothing went red.

Two recurring causes, both the same shapes reviewed on #69 and #71:

- **Fixtures their subject cannot act on.** `removeColor`'s fixture had no shapes and no problems, so two of the function's three parts ran zero iterations. `exchangeShapes` was only ever tested with the result clear of the swap. The movement cache was only ever given three entries against a 101-bucket table, so `moRehash` never ran.
- **Assertions adjacent to the claim.** `getMirrorTransform` checked only "non-zero", which any proper rotation satisfies. `getAttributePrefix` was read at index 0, where `index << 2` is 0 for any stride. The bitfield compared `"abc"` to `"ABC"`, which only says the two letter branches agree with each other.

### Corrected rather than covered

Three findings have no honest test-side fix. Rather than write an assertion that looks like coverage, the cases now say what is true:

- **`placementAllowed`'s `colorNumber() == 0` early return is unobservable.** Its two `bt_assert`s require both arguments `<= colorNumber()`, so with no colours the only legally passable pair is `(0, 0)` — which the `pc == 0 || res == 0` rule one line below allows anyway. The old comment claimed the case probed a pair that would be "DISALLOWED under the allow-list rule", which is the opposite of what the next case asserts.
- **The XML `illegal: <` branch is unreachable dead code.** `peekType` returns `LEGACY` only for `?` and `!`, so `parseLegacy`'s trailing `else` can never run; the input is rejected by `readName` instead. The case is retitled and now pins that message.
- **The declaration's `version` has no accessor.** The case promised it was "reported" but asserted only the encoding. Closing it properly needs a `getVersion()` on the parser, which is a production change, so it is retitled and the gap recorded.

### Deliberately not claimed

The new movement-cache case catches a rehash that **loses** entries, but not one that never grows the table — lookups chain, so a table stuck at 101 buckets still answers correctly. That is a slowdown, not a wrong answer, and no assertion here can see it. The case says so.

### Correction to an earlier version of this description

The first version of this PR claimed the rehash case killed a bucket-dropping mutant. It did not, and review caught it. `getMoValue` answers a miss by recomputing `moCalcValues` and re-inserting, and that function is pure, so a dropped entry is silently recomputed to the same value — comparing replayed values cannot detect a loss at all.

My verification was at fault, not only the test: the mutation harness restored a two-edit mutation in the wrong order, leaving an unrelated mutation in the tree, so the failure recorded against the rehash belonged to another case. (That also explains the single unreproducible test failure noted below — same contaminated binary, not flakiness.)

Fixed in `73f30379` by counting computations instead of comparing values, which is the only thing that can observe a miss. Dropping a bucket now fails the case.

### Also here

`screw()` and `someRotationMatches()` move from `test_voxeltable.cpp` into `test_helpers.h`, since the mirror case now needs the same chiral fixture and the same "is this really chiral?" guard.

`validCoordinate` and `onGrid` had no test at all, on any grid, despite fixtures throughout the suite justifying their cell choices by appealing to them.

### Verification

383 → **388 cases**, 23,603 → **38,161 assertions**, all passing. Clean under `--order rand` across six seeds. Every mutation above was reverted; no production file is touched by this branch.

Three defect classes from the PR reviews were checked against `master` and **do not** generalise: the sparse-grid vacuous fixtures (#69), the shape-id/part-index confusion (#71, killed by 6 and 3 cases), and writes outside a temp directory (#74, every write already uses RAII temp cleanup).


akoehn:
again by opencode union alpha:

 P2: Rehash test cannot detect lost entries (test/test_disassembly.cpp:1196–1205). getMoValue() recomputes missing entries, so dropping a bucket still produces identical replayed values. Use a counting override of moCalcValues() and assert replay performs no recomputation.

tomburns:
Confirmed, and fixed in `73f30379`. The model is right and I was wrong — worth being precise about how wrong, because I reported the opposite.

**The mechanism is exactly as stated.** `getMoValue` answers a miss by calling `moCalcValues` and re-inserting (movementcache.cpp:160-173), and `moCalcValues` is a pure function of the key. A dropped entry is therefore recomputed and returns byte-identical values. Comparing replayed values cannot detect a lost entry by construction — no amount of care with the fixture would have fixed that.

**My verification was faulty, not just my test.** I claimed in the PR description that a bucket-dropping mutant was killed. It was not. My mutation harness took per-file backups into a list, and for the one two-edit mutation (the bitfield letters, both branches of the same file) the second backup was taken *after* the first edit had been written — so restoring replayed the first mutation back in. The next mutation in the sequence ran against a tree that still had `bitfield.h` mutated, and the "killed by 1 case" I recorded for the rehash was the bitfield case failing. Re-run cleanly, dropping a bucket survived the entire 388-case suite.

That also explains something I flagged in the PR description as unexplained: a single test failure during committing that I could not reproduce in 15 runs. Same cause — the binary was built from the contaminated tree. No flakiness, just my harness.

**The fix is the one suggested.** A `CountingMovementCache` derived from `movementCache_c` with a counted `moCalcValues`; fill 144 keys past the 101-bucket threshold, then replay all of them and assert the counter has not moved. A replay that recomputes anything is a replay that found something missing.

One implementation note: `movementCache_0_c` can't be subclassed for this, because its `moCalcValues` is private and an override has no way to delegate to it. Deriving from the base instead turns out to be the better test anyway — the hashing, chaining and rehashing all live in `movementCache_c`, and the grid subclasses only supply geometry, so this puts the machinery under test on its own.

The counter is itself checked against a genuinely new key at the end, so "no recomputation" can't be satisfied by a counter that never increments.

| mutation | before | after |
|---|---|---|
| `moRehash` drops a bucket | survived | **killed** |
| `moRehash` never grows the table | survived | still survives — *see below* |

Still deliberately not claimed, and now stated in the case itself: a rehash that simply never grows the table. Lookups chain, so a table stuck at 101 buckets keeps answering correctly. That is a slowdown, not a wrong answer, and no assertion here can see it.

Stack rebased onto the updated #70; 388 cases, 38,158 assertions, green.

akoehn (dismissed):

PR #77, opened 2026-09-17
Replace hand-rolled hash tables with std unordered containers; bitfield C++20 hygiene

akoehn:
Replaces three hand-rolled hash tables (bucket vectors + intrusive `next` chains + manual rehash/`new`/`delete`) with `std::unordered_*` equivalents, and modernizes `bitfield_c` in place. Net −217 lines of plumbing. Plan and measurements: `design/2026-09-17-std-container-replacements.md`.

## Changes
- `voxelTable_c` → `unordered_multimap<hash, {index, transformation}>` (found to be GUI-only, not hot path)
- `movementCache_c` → `unordered_map<moKey, vector<uint>>`, original `moHashValue` kept as hasher
- `nodeHash`/`countingNodeHash` → `unordered_set` hashed/compared by node value; refcount discipline, `replaceNode`-on-shorter-way, and reverse-insertion scan order preserved; dead intrusive `disassemblerNode_c::next` removed
- `bitfield_c`: kept (no `std::bitset` swap — hex-literal tables need the custom codec), C++20 hygiene only: missing includes, `1ULL` shifts, `std::popcount`, const `notNull()`, defaulted copy assignment
- Not touched: `thread_c`, `Vector3D`, XML/gzip, domain types (rationale in design doc)

## Correctness
- `just test-all` 3/3 green, `just check` (cppcheck) clean
- Solver result counts bit-identical before/after on all benchmark puzzles (e.g. SolidSix: 588 asm / 179 sol / 4302868 iter)

## Performance (master `c2aab53` vs HEAD, both `-O3`, interleaved A/B pinned to one core, n=8)
| Puzzle | Before med | After med | Delta |
|---|---|---|---|
| SolidSixPieceBurrs | 8.78s | 8.96s | +2.1% |
| kangaroo | 1.98s | 2.04s | +3.0% |
| Simplicity | 3.54s | 3.72s | +5.2% |
| Excelsior | 2.23s | 2.25s | +1.1% |
Outliers on both sides, ranges overlap — perf-neutral within measurement noise. Per-run data in the design doc.


src/lib/bitfield.h
```diff
@@ -159,28 +168,13 @@ class bitfield_c {
...
-        s = (((s >> 2) & 0x3333333333333333ll) + (s & 0x3333333333333333ll));
-        s = (((s >> 4) + s) & 0x0f0f0f0f0f0f0f0fll);
-        s += (s >> 8);
-        s += (s >> 16);
-        s += (s >> 32);
-
-        res += (s & 0x3f);
-      }
+      for (int i = 0; i < ((bits+63)/64); i++)
+        res += std::popcount(field[i]);
```
tomburns, line 177: **`std::popcount` is not behaviour-preserving here — it fixes a latent undercount, and that changes the assembler.**

The old bithack finished with `res += (s & 0x3f)`. For a word with all 64 bits set the parallel count lands on `64`, and `64 & 0x3f == 0` — so a fully-set word contributed **0** instead of 64. `std::popcount` correctly returns 64.

That case is reachable: `src/lib/tabs_2/symmetries.inc` and `src/lib/tabs_2/unifiedsym.inc` both have entry **240** (the full symmetry group, `NUM_TRANSFORMATIONS_MIRROR == 240`) with word 0 == `0xFFFFFFFFFFFFFFFF`. So for a triangular-prism puzzle whose result shape and some part shape both carry the full symmetry group, `symmetries_2_c::countSymmetryIntersection()` (`src/lib/symmetries_2.cpp:114`) now returns a value 64 larger than before. That value feeds the symmetry-breaker selection in `src/lib/assembler_1.cpp:426-431` (`cnt < bestFound`), so a different `symBreakerShape` gets picked and the assembler's search order changes for those puzzles.

Correctness is still gated by `symmetriesLeft()`, so this is a fix rather than a regression — but the commit message and `design/2026-09-17-std-container-replacements.md` ("Phase 0 — implemented (obviously good, behaviour-preserving)") both claim no behaviour change, and `test/test_bitfield.cpp` has **no `countbits` case at all**. The benchmark corpus is cube-grid only, so nothing in this PR exercises the changed path.

Suggest: call it out as a fix in the doc, and add a `[bitfield]` case pinning `countbits()` on an all-ones word (e.g. set bits 0..63 and require 64).

src/lib/disassemblerhashes.h
```diff
@@ -36,14 +46,8 @@ class nodeHash {
...
-    /** current table size */
-    unsigned long tab_size;
-
-    /** current number of entries */
-    unsigned long tab_entries;
-
-    /** the hashtable */
-    std::vector<disassemblerNode_c*> tab;
+    /** the hashtable: node pointers hashed/compared by node value */
+    std::unordered_set<disassemblerNode_c*, disassemblerNodePtrHash, disassemblerNodePtrEqual> tab;
```
tomburns, line 50: **`nodeHash` was intrusive — this swap *adds* a heap allocation per stored node rather than removing one.**

The design doc says "the per-entry `hashNode` wrapper allocations are gone". That is true for `countingNodeHash` (which really did `new hashNode` per entry), but it is inverted for `nodeHash`: that table chained through `disassemblerNode_c::next`, so it allocated **nothing** per element. `std::unordered_set<disassemblerNode_c*>` now allocates one hash node per element (8-byte next pointer + 8-byte key, plus allocator overhead — ~32 bytes/node in practice). Deleting `next` recovers only 8 bytes, so the net is roughly +24 bytes per visited state.

This matters because `nodeHash` is the disassembler's largest structure: `src/lib/disassembler_0.cpp:80` keeps three fronts (`nodeHash closed[3]`) holding the entire visited set, and `disassemblernode.h` explicitly notes "as the tree can grow pretty large with a lot of nodes, it is important to keep the node small". Hard puzzles reach millions of nodes, where this is the difference between finishing and swapping.

The benchmark table in the design doc is consistent with a small real cost — every median moved the slow way (+1.1%, +2.1%, +3.0%, +5.2%), with no puzzle faster. Worth measuring peak RSS on one of the heavy corpus entries (e.g. `Jack Krijnen/Simplicity`, 5.1M iterations) before landing, and noting the allocation trade-off honestly in the doc.

src/lib/disassemblerhashes.cpp
```diff
@@ -36,99 +39,47 @@ nodeHash::~nodeHash(void) {
...
 
 const disassemblerNode_c * nodeHash::insert(disassemblerNode_c * n) {
 
-  unsigned long h = n->hash() % tab_size;
-
-  disassemblerNode_c * hn = tab[h];
-
-  while (hn) {
-    if (*hn == *n) {
+  auto it = tab.find(n);
```
tomburns: **Double lookup on the insert path of the disassembler's hottest loop.**

`tab.find(n)` followed by `tab.insert(n)` hashes the key twice and walks the bucket chain twice on every *miss* — and misses are the common case here (`disassembler_0.cpp` calls `closed[newFront].insert(st)` for every newly generated node). The old chained code walked the bucket exactly once. `insert()` already returns the existing element on collision, so one call does both jobs:

```suggestion
  auto [it, inserted] = tab.insert(n);

  if (!inserted) {
    disassemblerNode_c * hn = *it;
```

...with the matching `n->incRefCount();` kept and the trailing `tab.insert(n);` dropped. The same pattern applies to `countingNodeHash::insert()` below (`tab.find(n) != tab.end()` then `tab.insert(n)`), where `insert().second` gives the return value directly.

Given the recorded medians all moved 1-5% slower, reclaiming the redundant probe seems worth it.

design/2026-09-17-std-container-replacements.md
```diff
@@ -0,0 +1,181 @@
...
+No benchmark regressed; the heavy disassembler workload is within run-to-run
+noise. Correctness is pinned by `test_solver.cpp` exact assembly/solution
+counts plus the `[disasm][hash]`, `[voxeltable]` and `[movementcache]` cases.
+
+## Before/after timing (master `c2aab53` vs branch HEAD, `-O3` both)
+
+Build flags verified in `compile_commands.json`: `buildtype=release`,
+`-O3`, no `-DNDEBUG` (so `bt_assert` is active in both builds — equal
+footing). Method: both `burrTxt` binaries kept side by side, 8 interleaved
+runs each pinned to one core (`bench/bench_solve.py --ab ... --cpu 7`).
```
tomburns, line 122: **The benchmark harness this PR's justification rests on is not in the repo.**

`bench/bench_solve.py` does not exist anywhere in the tree (checked at this commit), and the raw sweep data is cited as `/tmp/opencode/btfiles_sweep.csv` (line 148) — a scratch path that will be gone. Since "perf-neutral within measurement noise" is the stated basis for accepting a 1-5% median slowdown on every measured puzzle, nobody can re-run or extend the measurement later, and the numbers cannot be rechecked if the disassembler memory profile is questioned.

Either commit the harness under `bench/` (it is small and the corpus path is already covered by the `puzzles/` entry in `.gitignore`) or drop the precise per-run tables and state the methodology in prose.

akoehn:
All four comments addressed in 20ad9a7 + e4a3644 (test-all 3/3, cppcheck clean):

1. popcount undercount — confirmed, and worse than cosmetic: entry 240 of symmetries.inc is all-ones, so countSymmetryIntersection was 64 short for full-symmetry shapes, feeding the assembler_1 symBreaker choice. Doc now calls it a fix (not behavior-preserving), test added for full-word counts.
2. nodeHash allocation — confirmed inverted for nodeHash (intrusive was 0/entry; unordered_set ~32B). Measured peak RSS on Simplicity: 452.7MB both before/after (assembler-matrix-dominated, per-entry delta doesn't move the process peak). Trade-off documented honestly.
3. Double lookup — fixed in both inserts via the insert() return value; re-benched (Simplicity +1.3%, SolidSix +0.2%, kangaroo +7.6% on ~2s scale).
4. Bench harness — bench/bench_solve.py + bench/sweep_btfiles.py committed (not wired into CI); doc references updated.

src/lib/bitfield.h
tomburns: Fixed in 20ad9a7 / e4a3644 — countbits() case added pinning the all-ones word to 64, and the doc's Phase 0 section now calls this out as a fix rather than behaviour-preserving.

src/lib/disassemblerhashes.h
tomburns: Addressed in e4a3644 — measured whole-process peak RSS on Simplicity (452.7 MB before/after, no move), documented the +~24 bytes/state estimate for the disassembler tables, and accepted the trade-off for the RAII/exception-safety win.

src/lib/disassemblerhashes.cpp
tomburns: Fixed in 20ad9a7 exactly as suggested — both nodeHash::insert and countingNodeHash::insert now do a single insert() call.

design/2026-09-17-std-container-replacements.md
tomburns: Fixed in e4a3644 — bench/bench_solve.py and bench/sweep_btfiles.py are committed, and the doc's per-run tables now cite that harness instead of the /tmp path.

tomburns (approved):

PR #83, opened 2026-09-19
feat(solver): extend SIMD exact cover up to 32768 cols, AVX-512 & 2.7x disasm kernel

akoehn:
## Overview

Stacked on #82 (`assembler-simd`).

This PR extends SIMD acceleration across both the assembler and disassembler engines:

1. **SIMD Exact Cover up to 32,768 Columns**: Generalizes `SimdBitset<N_WORDS>` (aligned to 64 bytes) with compile-time power-of-2 specializations (256, 512, 1024, 2048, 4096, 8192, 16384, and 32768 columns) with compile-time unrolled vector loops and automatic solver selection in `assembler_0_c`. Supports large puzzle grids up to 32³ voxels (including 18³ = 5,832 voxels via the 8,192-column tier).
2. **Planar Transitive Closure & AVX2 / NEON**: Reorganizes movement constraint matrices into an aligned planar buffer per direction, enabling vectorization of Roy-Floyd-Warshall all-pairs shortest paths via `_mm256_min_epu32` / `_mm256_add_epi32` (and ARM NEON `vminq_u32` / `vaddq_u32`).
3. **Flat 64-byte Cache-Line Movement Cache**: Replaces node-based `std::unordered_map` with an open-addressing table where each `moEntry` fits in exactly one 64-byte L1 cache line, eliminating pointer chasing, node allocations, and `std::vector` heap buffers.
4. **AVX-512 Vectorization**: Adds specialized AVX-512 kernels (`_mm512_test_epi64_mask` in exact cover filtering and `_mm512_min_epu32` in `closureFull`) with runtime feature toggles (`BURRTOOLS_NO_AVX512=1`).

---

### Disassembly & Exact Cover (Interleaved A/B, 5 runs):

| Puzzle | Before | After | Speedup | Peak RSS Delta |
| `Excelsior.xmpuzzle` | 2.43s | 0.90s | **2.71×** | +0.17 MB (+1.7%) |
| `kangaroo.xmpuzzle` | 1.66s | 0.81s | **2.04×** | +0.18 MB (+1.1%) |

### Linux `perf` Profile on `Excelsior`:

- **Total CPU cycles**: 4.87 billion → 2.74 billion (**-43.8% total CPU cycles eliminated**).
- `closureFull()`: 1.93 billion cycles → 410 million cycles (**4.7× faster**).
- `movementCache` lookups: `std::_Hashtable` is completely eliminated (0.0%), reducing lookup overhead from 890M cycles to 399M cycles (**2.2× faster**).

---

## Verification

- `just test` & `just test-all` pass (all fast and slow regression tests pass).
- `just check` (cppcheck static analysis) passes 100% cleanly.
- `just build-win` (MinGW64 Windows cross-compilation) succeeds with zero warnings/errors.


src/lib/assembler_0.cpp
```diff
@@ -1804,7 +1804,7 @@ bool assembler_0_c::canUseSimd(void) const {
 
   int res_filled = getResultShape(problem)->countState(voxel_c::VX_FILLED);
   unsigned int max_col = piecenumber + res_filled;
-  if (max_col > 512)
+  if (max_col > 32768)
```
tomburns: Raising this gate from 512 to 32768 admits problems whose SIMD footprint and per-node cost are ~64x larger, and nothing else in `canUseSimd` compensates.

Two concrete consequences:

1. **Memory.** `SimdExactCover<BitsetType>::rows` stores one `BitsetType` per row by value. At the 32768 tier `sizeof(Row)` is ~4160 bytes (4 KB mask + ids + a `std::vector`, rounded to the 64-byte alignment). A puzzle at, say, 20000 columns typically has O(10^5) placements, so `rows` alone is ~400 MB-4 GB. The gate only looks at `piecenumber + res_filled`, never at the row count, so a puzzle that used to fall back to DLX now allocates until the process dies or thrashes.

2. **Per-node cost.** `search()` does `std::fill(ctx.col_counts.begin(), ctx.col_counts.end(), 0)` at *every* visited node, and `col_counts` is sized `num_columns`. At 32768 columns that is 32768 stores per search node before any real work, plus the MRV scan over `active_column_list`. Combined with the 4 KB `operator|` temporary per level, it is not obvious that the SIMD path still beats DLX above a few thousand columns — it may well be a large regression exactly where this change makes it eligible.

AGENTS.md section 7 requires optimisations to be validated on the curated 10-puzzle corpus. The PR body reports only `Excelsior` and `kangaroo`, both 6-piece puzzles well under 512 columns, so neither of them exercises any tier this change adds. Could you show corpus numbers (or at least one puzzle in the 4096+ range) comparing the new tiers against the DLX fallback? If they lose, the right fix is to lower this gate back to whatever tier is actually a win, and/or add a `rows * sizeof(BitsetType)` budget check.

src/lib/simd_exact_cover.h
```diff
@@ -36,237 +36,148 @@
...
 };
 
+using SimdBitset256 = SimdBitset<4>;
+using SimdBitset512 = SimdBitset<8>;
+using SimdBitset1024 = SimdBitset<16>;
+using SimdBitset2048 = SimdBitset<32>;
+using SimdBitset4096 = SimdBitset<64>;
+using SimdBitset8192 = SimdBitset<128>;
+using SimdBitset16384 = SimdBitset<256>;
+using SimdBitset32768 = SimdBitset<512>;
```
tomburns, line 123: At the 32768 tier a `SimdBitset` is 4096 bytes, and the search recursion keeps one alive per level.

In `simd_exact_cover.cpp` the recursive step is

```cpp
search(depth + 1, occupied | candidate_row.mask, ctx, callback, abort_flag, iterations);
```

`operator|` returns a `BitsetType` **by value**; the temporary is bound to the callee's `const BitsetType &occupied` and lives, on the caller's stack frame, for the whole duration of the recursive call. So peak stack is roughly `max_depth * sizeof(BitsetType)`, and `max_depth == num_pieces`.

Failure scenario: a puzzle with ~10000 filled voxels and 128 pieces selects `SimdExactCover32768`; the search recurses to depth 128, holding 128 x 4 KB = 512 KB of bitset temporaries. `std::thread` workers on macOS get a 512 KB stack by default, so this is a stack overflow (silent crash, not an assertion) in the parallel solver added in #79/#81. Even on Linux's 8 MB default, a 500-piece puzzle (still well under the 32768 column cap) would need 2 MB per worker thread. At the old 512-column cap the same expression cost 64 bytes/level and this was a non-issue.

Simplest fix is to make `occupied` a per-depth slot in `SearchContext` (push/pop into `ctx`, heap-backed) rather than a stack temporary chain.

tomburns (commented):
Review of `assembler-simd..more-simd` only.

The generalisation of `SimdBitset256`/`SimdBitset512` into `SimdBitset<N_WORDS>` is a clean refactor and the AVX-512 additions are correctly guarded by runtime `__builtin_cpu_supports` with an AVX-512 -> AVX2 -> scalar/NEON fallback chain, so there is no SIGILL hazard and the arm64 build is unaffected (the whole x86 block is `#if`-ed out). The `closureFull` planar rewrite is semantically faithful to the loop it replaces: directions are independent in the original strided loop, `_mm256_min_epu32` matches the unsigned `sum < row_y[x]` test, and `row_y`/`row_k` never alias because `y == k` is skipped. Good catch on `NUM_WORDS` vs `sizeof(BitsetType)` now that every bitset is `alignas(64)`.

My concerns are concentrated in two places: (a) raising the column cap from 512 to 32768 changes the per-row and per-search-node cost by 64x with no guard and no benchmark evidence at the new tiers, and (b) the claimed 2.7x is attributed in the title to the vector kernels, but both benchmark puzzles have 6 pieces, so no vector instruction in the new disassembly kernel ever executes on them. Details inline.

Two smaller process notes not worth their own inline threads: the new `BURRTOOLS_NO_AVX512` toggle is not added to the env-var table in AGENTS.md section 4, which that section requires for A/B toggles; and `simdDisabled()` in movementanalysator.cpp makes `BURRTOOLS_NO_SIMD=1` also switch the disassembler kernel, which conflates assembler and disassembler in any `bench_solve.py --ab` run that uses that toggle as its baseline.

I did not build or run anything (concurrent reviewers), so the compile-time and performance points are reasoned from the source, not measured.

src/lib/simd_exact_cover.cpp
```diff
@@ -82,33 +88,67 @@ void SimdExactCover<BitsetType>::registerNodeAlias(unsigned int node_id, uint32_
...
+void SimdExactCover<BitsetType>::filterRowsAvx512(
   const std::vector<uint32_t> &src,
   const BitsetType &chosen_mask,
   std::vector<uint32_t> &dst
 ) const {
-  if constexpr (sizeof(BitsetType) == 32) {
-    __m256i va = _mm256_load_si256(reinterpret_cast<const __m256i*>(chosen_mask.words));
+  if constexpr (BitsetType::NUM_WORDS >= 8) {
+    constexpr size_t N_VEC = BitsetType::NUM_WORDS / 8;
+    __m512i va[N_VEC];
```
tomburns, line 105: Hoisting the whole `chosen_mask` into a `__m512i va[N_VEC]` array before the row loop is a pessimisation at the large tiers, and the same pattern was introduced in `filterRowsAvx2` (`__m256i va[N_VEC]`) and `filterRowsNeon` (`uint64x2_t va[N_VEC]`).

For `SimdBitset32768`, `N_VEC` is 64 here (128 for AVX2, 256 for NEON) — i.e. a 4 KB array. There are only 32 zmm / 16 ymm / 32 neon registers, so the compiler cannot keep it in registers: this compiles to a mandatory 4 KB stack allocation plus a full 4 KB copy of `chosen_mask` on **every** `filterRows` call, i.e. once per search node. Meanwhile the inner loop `break`s on the first overlapping 512-bit chunk, which for a sparse exact-cover instance is almost always `i == 0` or `i == 1`, so the other 63 preloaded vectors are never read.

Reading `chosen_mask.words[i * 8]` directly inside the loop is both simpler and strictly cheaper here (the value is loop-invariant and the L1 line is hot). The preload is only a win for the 256/512-bit tiers where `N_VEC` is 1-2 and the vectors genuinely stay in registers, which is presumably where it came from.

src/lib/movementcache.h
```diff
@@ -65,12 +64,21 @@ class movementCache_c {
...
 
-  struct moKeyHash {
-    size_t operator()(const moKey & k) const noexcept;
+  /**
+   * 64-byte cache-line aligned entry: key + values fit in a single L1 cache line,
+   * eliminating pointer chasing and heap node allocations.
+   */
+  struct alignas(64) moEntry {
+    moKey key;
+    std::array<unsigned int, 8> values{};
```
tomburns, line 73: `values` is hard-coded to 8, but the number of values written and read is `numDirections()`, a pure virtual supplied by each grid-specific subclass. Today `movementCache_0_c` returns 3 and `movementCache_1_c` returns 4, so this happens to fit — but nothing enforces it, and both failure modes are silent:

- the store loop in `getMoValue` is `for (size_t i = 0; i < vals.size() && i < 8; ++i)`, so a grid with more than 8 directions silently **drops** the extra values;
- `memcpy(movements, moTable[idx].values.data(), numDirections() * sizeof(unsigned int))` then reads past the 32-byte array into `occupied`/padding/the next entry, so the caller gets garbage movement distances and the disassembler produces a wrong (or wrongly-levelled) disassembly with no diagnostic.

The old `std::vector<unsigned int>` sized itself from `moCalcValues`, so this constraint is new. Please add a `bt_assert(numDirections() <= 8)` in the `movementCache_c` constructor (or a `static constexpr size_t MAX_DIRECTIONS = 8` used by both the array and the assert) so a future grid type fails loudly instead of returning wrong data. Worth noting this is exactly the kind of cube-grid-shaped assumption that would bite whoever adds a rhombic or sphere movement cache.

src/lib/movementcache.cpp
```diff
@@ -100,21 +98,53 @@ const voxel_c * movementCache_c::getTransformedShape(unsigned int s, unsigned ch
...
 {
   /* find out the shapes that the pieces have */
   unsigned int s1 = pieces[p1];
   unsigned int s2 = pieces[p2];
 
   moKey key{s1, s2, dx, dy, dz, t1, t2};
+  unsigned int h = moHashValue(s1, s2, dx, dy, dz, t1, t2);
+
+  size_t idx = h & moMask;
+  while (moTable[idx].occupied) {
```
tomburns, line 128: Switching from `std::unordered_map` to open addressing also switches the bucket derivation from *modulo a prime* to *mask off the low bits*, and `moHashValue` is not strong enough in its low bits for that.

Taking the multiplier constants mod 1024: `dx` contributes `16*dx`, `dy` contributes `20*dy`, `s1` contributes `516*s1`, `s2` `642*s2`, `t1` `26*t1` — all even, and `dx` in particular can only land on 64 of the 1024 slots. Only `dz` (337) and `t2` (1) are odd. With libstdc++'s prime-modulo bucketing the high bits were folded back in, which is why this hash was adequate before; `h & moMask` throws them away.

Failure scenario: a burr where most queried moves are along one axis with the other two offsets pinned (`dy == dz == 0`, `t1 == t2 == 0` for the identity-oriented pieces) — the bucket index collapses to `(16*dx + 516*s1 + 642*s2) & 1023`, a small even subset. Under linear probing those keys form long contiguous runs, so lookups that should be O(1) degrade toward O(table size), and the 70% load factor makes it worse. That is a latent throughput cliff on exactly the puzzle shapes this PR is optimising for.

A final mixing step before the mask fixes it cheaply, e.g. `idx = (h * 2654435769u) >> (32 - log2(cap))`, or a murmur/splitmix finalizer on `val` at the end of `moHashValue`.

src/lib/movementanalysator.cpp
```diff
@@ -167,23 +291,52 @@ void movementAnalysator_c::closureFull(void) {
...
+      const unsigned int * src = matrix.data() + (size_t)y * rowStep + d;
+      unsigned int * dst = planar_block.data() + (size_t)y * n;
+      for (unsigned int x = 0; x < n; x++) {
+        dst[x] = src[x * dirs];
+      }
+    }
+
+    // 2. Transitive closure on contiguous planar block
+#if (defined(__x86_64__) || defined(_M_X64)) && (defined(__GNUC__) || defined(__clang__))
+    if (has_avx512 && n >= 16) {
```
tomburns, line 311: `n` here is `pieces->size()`, the number of pieces in the current disassembly sub-problem. Both puzzles in the PR's benchmark table are 6-piece puzzles (`bench/run_suite.sh` describes `kangaroo` and `Excelsior` as "6 unique pieces"), so on the numbers quoted:

- `has_avx512 && n >= 16` is false, so `rfw_avx512` never runs;
- control falls to `rfw_avx2`, whose only vector loop is `for (; x + 8 <= n; x += 8)` — with `n == 6` the body never executes, and every element goes through the scalar tail.

So the measured 2.71x / 2.04x and the "closureFull 4.7x faster" perf number were produced with **zero** vector instructions executed in this kernel. The win is real but it comes from the planar pack/unpack (turning a stride-`dirs` scalar walk into a contiguous one) plus the flat movement cache — not from AVX2 or AVX-512, which the PR title attributes it to. Please either re-measure on a puzzle with n >= 8 (and n >= 16 for the AVX-512 kernel) or restate the claim, otherwise the stack carries an unvalidated attribution forward into #84.

Relatedly: `rfw_avx512` and the vector body of `rfw_avx2` are, on this corpus, untested dead code. There is no test that runs a disassembly twice — once with `BURRTOOLS_NO_SIMD=1` and once without, or against the `BURRTOOLS_NO_DISASM_OPT` legacy path — and asserts identical `movesText()`. Since the four kernels here are hand-written duplicates of one algorithm, that equivalence test is the thing that would catch a divergence between them.

test/test_simd_exact_cover.cpp
```diff
@@ -131,6 +131,222 @@ TEST_CASE("SimdBitset512 operations", "[simd][bitset]") {
...
+  disjoint_set.set(1);
+  disjoint_set.set(10000);
+  REQUIRE(is_disjoint_scalar(b, disjoint_set));
+
+#if (defined(__x86_64__) || defined(_M_X64)) && (defined(__GNUC__) || defined(__clang__))
+  if (__builtin_cpu_supports("avx2")) {
+    REQUIRE(is_disjoint_avx2(b, disjoint_set));
+    REQUIRE_FALSE(is_disjoint_avx2(b, target));
+  }
+  if (__builtin_cpu_supports("avx512f")) {
```
tomburns, line 298: Every AVX-512 assertion in this file is inside `if (__builtin_cpu_supports("avx512f"))`, which means on the two machines this project is actually built on they assert nothing: macOS arm64 (the stated dev platform, where the entire `#if defined(__x86_64__)` block is not even compiled) and GitHub's `ubuntu-latest` runners, which are not guaranteed to expose AVX-512 and on the current AMD/Intel fleet frequently do not. A Catch2 run that skips a path reports as green, so an AVX-512-only bug would ship silently.

Also, what is covered here is only the free functions `is_disjoint_avx512` / `is_disjoint_avx2`. The code that actually runs in the solver is `SimdExactCover::filterRowsAvx512` and `filterRowsAvx2` (plus `rfw_avx512` / `rfw_avx2` in movementanalysator), and none of those have any test at all — including the `if constexpr (NUM_WORDS >= 8) ... else filterRowsAvx2(...)` fallback branch.

The testable design is already there: `use_avx2` / `use_avx512` are set from `BURRTOOLS_NO_AVX2` / `BURRTOOLS_NO_AVX512` in the constructor. Exposing them as constructor parameters (or a setter) would let one test solve the same instance three ways — forced scalar, forced AVX2, forced AVX-512 — and `REQUIRE` bit-identical solution sets, which is the check that makes the ISA dispatch trustworthy. That equivalence test would run meaningfully on arm64 too (scalar vs NEON).

test/test_simd_exact_cover.cpp
```diff
@@ -131,6 +131,222 @@ TEST_CASE("SimdBitset512 operations", "[simd][bitset]") {
...
+    REQUIRE(is_disjoint_avx512(b, disjoint_set));
+    REQUIRE_FALSE(is_disjoint_avx512(b, target));
+  }
+#endif
+}
+
+TEST_CASE("SimdExactCover extended solve up to 32768", "[simd][exact_cover]") {
+  for (unsigned int offset : {600u, 1200u, 3000u, 6000u, 12000u, 24000u}) {
```
tomburns: This is a good breadth test across tiers, but it does not touch any of the boundaries where the new capacity logic can go wrong.

- The largest column index used is `24000 + 6 = 24006`; nothing in the suite solves with a bit set at 32767, the top of the `SimdBitset32768` tier. Given that `setRequiredColumn`/`addRow` guard on `col < BitsetType::NUM_WORDS * 64` — i.e. against *capacity*, not against the constructor's `num_columns` — the last word of the largest bitset is the one place an off-by-one would live, and it is unexercised.
- No case sits on a tier boundary. `createSimdSolver` switches at `max_col <= 256 / <= 512 / ... / <= 16384`, so `max_col == 16384` and `max_col == 16385` select different bitset widths; `max_col == 32768` is the last accepted value and 32769 must fall back to DLX via `canUseSimd`. None of those transitions are asserted anywhere, here or in the assembler tests.
- Sizes are all multiples of 1000 plus 7, so every instance sits mid-word. A size like `num_columns == 32768` (exactly 512 words) versus `32705` (one bit into the last word) versus `32767` would be the cases worth adding, since `col_counts` is sized `num_columns` while the bitset is sized to capacity and the two diverge by up to 2x at every tier.

A cheap addition: for each tier, build the instance with `num_columns` equal to the tier's exact capacity and place the required columns at the top of the range.

src/lib/movementanalysator.cpp
```diff
@@ -110,11 +110,135 @@ void movementAnalysator_c::prepareFill(void) {
...
 }
 
+static bool simdDisabled() {
+  static const bool disabled = (std::getenv("BURRTOOLS_NO_SIMD") != nullptr || std::getenv("BURRTOOLS_NO_AVX2") != nullptr);
+  return disabled;
+}
+
+#if (defined(__x86_64__) || defined(_M_X64)) && (defined(__GNUC__) || defined(__clang__))
+#pragma GCC push_options
+#pragma GCC target("avx512f")
```
tomburns: **Blocking (portability): this file does not compile under Clang on x86-64, and this PR makes the breakage substantially worse.**

`#pragma GCC push_options` / `#pragma GCC target(...)` is a GCC extension that **Clang does not implement** — it emits `-Wunknown-pragmas` and ignores the directive, after which every intrinsic in the block fails to compile because the target feature was never enabled.

Verified with Apple Clang 21.0.0 targeting x86-64, compiling this branch's real sources:

```
$ clang++ -std=c++20 -arch x86_64 -Isrc -Isrc/lib -c src/lib/movementanalysator.cpp -o /dev/null
20 errors
```

The error is the usual shape:

```
error: always_inline function '_mm256_load_si256' requires target feature 'avx',
  but would be inlined into function 'filterRowsAvx2' that is compiled without
  support for 'avx'
error: AVX vector return of type '__m256i' without 'avx' enabled changes the ABI
```

Comparing the two branches, per translation unit (x86-64 Clang, error count):

| file | on #82 | on #83 |
|---|---|---|
| `src/lib/movementanalysator.cpp` | **0** | **20** |
| `src/lib/simd_exact_cover.cpp` | 6 | 6 |
| `src/lib/simd_huang_cover.cpp` | 6 | 6 |
| `test/test_simd_exact_cover.cpp` | 1 | 1 |

So the pragma pattern is inherited from #82, but #83 is what brings `movementanalysator.cpp` into the broken set — this is the single largest new contribution to the breakage in the stack.

**No CI job can see this.** Linux and Windows build with GCC (where the pragma works); `macos-latest` is arm64, where this whole block is `#if`-excluded. An x86-64 Clang build — Linux/clang, FreeBSD, or an Intel Mac — fails outright. Note also that the three ignored-pragma *warnings* alone would fail `just build-werror` before the hard errors are reached.

**Confirmed fix:** replace the pragma pair with a per-function attribute, which both GCC and Clang honour. Verified it compiles clean for x86-64 and emits the intended instruction:

```
$ clang++ -std=c++20 -arch x86_64 -c probe.cpp     # __attribute__((target("avx2")))
0 errors
$ clang++ -std=c++20 -arch x86_64 -S probe.cpp -o - | grep -c vptest
1
```

So `__attribute__((target("avx512f")))` / `__attribute__((target("avx2")))` on each function is a drop-in, and the `push_options`/`pop_options` wrappers can be deleted. Worth fixing in #82 and rebasing rather than patching twice, since the same change is needed in `simd_exact_cover` and `simd_huang_cover`.

Adding an x86-64 Clang compile job would keep this from recurring; it is cheap (`-fsyntax-only` is not enough — it must reach codegen).

src/lib/assembler_0.cpp
akoehn: Fixed in `bbf9528`. Gated `canUseSimd()` in `assembler_0.cpp` to `max_col <= 2048` to avoid memory explosion for large column counts while preserving extended template tiers for unit testing.

src/lib/simd_exact_cover.h
akoehn: Fixed in `bbf9528`. Capped `canUseSimd()` in `assembler_0.cpp` at 2048 columns.

src/lib/simd_exact_cover.cpp
akoehn: Addressed in `bbf9528`.

src/lib/movementcache.h
akoehn: Addressed in `bbf9528`.

src/lib/movementcache.cpp
akoehn: Addressed in `bbf9528`.

src/lib/movementanalysator.cpp
akoehn: Fixed in `bbf9528`. Replaced `#pragma GCC target` with `__attribute__((target(...)))` and decoupled disassembler SIMD via `BURRTOOLS_NO_DISASM_SIMD`.

test/test_simd_exact_cover.cpp
akoehn: Fixed in `bbf9528`. Added exact-capacity boundary tests across all tiers (256 through 32768).

test/test_simd_exact_cover.cpp
akoehn: Fixed in `bbf9528`. Added differential test for `movementAnalysator_c` on 9-piece Prisgon with and without `BURRTOOLS_NO_DISASM_SIMD`.

src/lib/movementanalysator.cpp
akoehn: Fixed in `bbf9528`. Replaced `#pragma GCC target` with `__attribute__((target(...)))` across `movementanalysator.cpp`.

akoehn:
All review comments have been addressed in commit bbf9528:
- Capped `canUseSimd()` in `assembler_0.cpp` at 2048 columns to prevent excessive allocations for large column counts while keeping the extended template tiers available for unit testing.
- Decoupled disassembler SIMD benchmarking via `BURRTOOLS_NO_DISASM_SIMD` so assembler and disassembler toggles do not conflate.
- Replaced `#pragma GCC target` with `__attribute__((target(...)))` across `movementanalysator.cpp`.
- Documented `BURRTOOLS_NO_AVX512` and `BURRTOOLS_NO_DISASM_SIMD` in `AGENTS.md`.
- Added capacity boundary tests across all tiers in `test/test_simd_exact_cover.cpp`.
- Added differential test for `movementAnalysator_c` on 9-piece Prisgon with and without `BURRTOOLS_NO_DISASM_SIMD` in `test/test_disassembly.cpp`.

src/lib/movementcache.cpp
tomburns: No finalizer was added — `movementcache.cpp:127` is still

```cpp
  size_t idx = h & moMask;
```

taking the raw `moHashValue` low bits, and `moHashValue` is unchanged. So the analysis in the original comment still holds exactly: mod 1024 the multipliers give `16*dx`, `20*dy`, `516*s1`, `642*s2`, `26*t1` — all even, with `dx` reaching only 64 of 1024 slots — and only `dz` (337) and `t2` (1) are odd. Under the old prime-modulo bucketing the high bits folded back in; `h & moMask` discards them, and with linear probing at a 70% load factor the clustered keys form long runs.

One line, at the mask site:

```cpp
  // Fibonacci finalizer: moHashValue's low bits are weak (most multipliers are
  // even), and masking discards the high bits that prime-modulo bucketing used.
  size_t idx = (static_cast<size_t>(h) * 2654435769u) >> (32 - moBits);
```

(or a splitmix/murmur finalizer on `val` at the end of `moHashValue`, if you'd rather keep the fix at the source). Either way it's a couple of cycles against a latent throughput cliff on exactly the burr shapes this PR targets.

src/lib/movementanalysator.cpp
tomburns: The "Relatedly" half landed, and well — `test/test_disassembly.cpp:1510`, "movement analysator: differential comparison with and without `BURRTOOLS_NO_DISASM_SIMD` on 9-piece Prisgon", is exactly the equivalence test this asked for, on a puzzle with `n >= 8` so the AVX2 vector body actually executes. Thanks.

The attribution half is unchanged — the PR description still reports 2.71× / 2.04× and "closureFull 4.7× faster" without qualification.

**But I should flag that my premise here looks wrong**, and if it is, there's nothing left to restate. I took `n` to be 6 from `bench/run_suite.sh` describing the puzzles as "6 unique pieces" — but `n` is `pieces->size()`, the piece count in the subproblem, not the count of distinct shapes. This PR's own design doc says so directly:

- `design/2026-09-19-disassembly-simd-bitboards.md:47` — "for `Excelsior` $N = 18$, 6-piece burrs $N = 6$, `kangaroo` $N = 36$"
- `:142` — "For $N = 18$ (as in `Excelsior`), 18 elements is just **three 256-bit AVX2 vectors**"

At N=18 and N=36 both `rfw_avx2`'s `x + 8 <= n` loop and the `has_avx512 && n >= 16` gate are satisfied, so the kernels do run and the attribution is fine as written.

I can't verify it myself — `Excelsior.xmpuzzle` isn't in `examples/`. Could you confirm the piece counts for the two benchmark puzzles? If they're 18 and 36, close this; if either is really 6, the restatement is still owed.

test/test_simd_exact_cover.cpp
tomburns: The capacity-boundary tests referenced in the reply are the answer to the *other* test thread (the tier-coverage one); this one is about ISA dispatch, and it's untouched.

`BURRTOOLS_NO_AVX2` and `BURRTOOLS_NO_AVX512` still appear nowhere under `test/` on any branch in this stack. So:

- every AVX-512 assertion is still inside `if (__builtin_cpu_supports("avx512f"))` and still asserts nothing on macOS arm64 or on GitHub's `ubuntu-latest` fleet — a green Catch2 run that skipped the path;
- the code that actually runs in the solver — `SimdExactCover::filterRowsAvx512` / `filterRowsAvx2` / `filterRowsNeon`, plus the `if constexpr (NUM_WORDS >= 8)` fallback branch — still has no test at all.

The design for this is already in place, and #83 added the last missing piece: `SimdExactCover`'s constructor now sets `use_avx2` / `use_avx512` / `use_neon` from env vars. Exposing them as constructor parameters or a setter lets one test solve the same instance three ways and `REQUIRE` identical solution sets:

```cpp
for (auto isa : {Isa::Scalar, Isa::Avx2, Isa::Avx512}) {
  if (!isaAvailable(isa)) continue;
  auto solver = makeSolver(cols, pieces, isa);
  ... // same rows
  CHECK(solutionsFor(isa) == scalarSolutions);
}
```

and the same shape runs meaningfully on arm64 as scalar-vs-NEON. `ScopedEnv` in `test/test_disassembly.cpp` is a ready-made alternative if you'd rather drive it through the environment than widen the constructor.

src/lib/simd_exact_cover.cpp
tomburns: Unchanged — the preload arrays are still hoisted in all three kernels:

- `simd_exact_cover.cpp:105` — `__m512i va[N_VEC];` then `va[i] = _mm512_load_si512(&chosen_mask.words[i * 8]);`
- `:137` — `__m256i va[N_VEC];` / `_mm256_load_si256(&chosen_mask.words[i * 4]);`
- `:165` — `uint64x2_t va[N_VEC];`

At the 32768 tier that is a 4 KB stack array plus a full 4 KB copy of `chosen_mask` on every `filterRows` call — once per search node — while the inner loop still `break`s on the first overlapping chunk, so for a sparse instance almost all of it is never read.

The fix is to drop the array and read the loop-invariant operand inline, e.g. for the AVX-512 kernel:

```cpp
for (uint32_t idx : src) {
  ...
  for (unsigned int i = 0; i < N_VEC; i++) {
    __m512i va = _mm512_load_si512(reinterpret_cast<const void*>(&chosen_mask.words[i * 8]));
    __m512i vb = _mm512_load_si512(reinterpret_cast<const void*>(&cand.mask.words[i * 8]));
    if (_mm512_test_epi64_mask(va, vb)) { overlap = true; break; }
  }
}
```

(the L1 line is hot, so the reload is free in practice). If the preload is worth keeping for the 256/512 tiers where `N_VEC` is 1-2 and the vectors genuinely stay in registers, gate it on `if constexpr (N_VEC <= 2)` rather than applying it at every width.

src/lib/movementcache.h
tomburns: Still unguarded on `bbf9528` — the only `bt_assert` added anywhere in `movementcache.{h,cpp}` is the pre-existing `bt_assert2(sh->transform(t))` at `movementcache.cpp:92`. The 8 is still a bare literal, and both failure modes are still silent: `getMoValue`'s store loop caps at 8 and drops the rest, and the read side does `memcpy(movements, moTable[idx].values.data(), numDirections() * sizeof(unsigned int))` straight past the array into `occupied` and the next entry.

```suggestion
    static constexpr size_t MAX_DIRECTIONS = 8;
    std::array<unsigned int, MAX_DIRECTIONS> values{};
```

plus, in the `movementCache_c` constructor:

```cpp
bt_assert(numDirections() <= moEntry::MAX_DIRECTIONS);
```

so a future grid type (rhombic, sphere) fails loudly at construction instead of returning wrong movement distances and a silently wrong disassembly level.

src/lib/movementcache.cpp
tomburns: **Severity: minor — latent throughput risk, not correctness.** Lookups still return the right answer; the concern is clustering under linear probing on a specific query shape, and there's no measurement showing it biting on the current corpus. Not a merge blocker — resolving.

Keeping the one-line finalizer in the thread above for whenever the movement cache next shows up in a profile.

test/test_simd_exact_cover.cpp
tomburns: **Severity: minor — test-coverage gap, not a defect.** Resolving so it doesn't gate the merge.

Recording it as a follow-up rather than dropping it, because the gap is structural: the AVX-512 assertions are skipped on both machines this project is actually built on, so that path stays green-by-omission until the forced-ISA equivalence test exists. It pairs naturally with the `BURRTOOLS_NO_SIMD` suite parameterisation tracked on #82.

src/lib/simd_exact_cover.cpp
tomburns: **Severity: minor — not a merge blocker**, and smaller now than when filed. #83 caps `assembler_0_c::canUseSimd()` at 2048 columns, so the production path tops out at `SimdBitset2048` (32 words → `N_VEC` of 4 for AVX-512, 8 for AVX2), not the 64/128/256 that made this a 4 KB stack copy. The large tiers are reachable only from the unit tests now.

So this is a micro-optimisation on a path that is no longer hot, not a defect. Resolving; the `if constexpr (N_VEC <= 2)` gating above is worth doing if the cap is ever raised again.

src/lib/movementcache.h
tomburns: **Severity: minor — defensive, currently unreachable.** `movementCache_0_c` returns 3 and `movementCache_1_c` returns 4, so nothing today exceeds the array. This is a guard against a future grid type, not a live bug. Not a merge blocker — resolving.

The assert is still worth adding whenever this file is next touched, since both failure modes are silent (dropped values on write, a read past the array into `occupied` on the way out) and would surface as a wrong disassembly level rather than a crash.

src/lib/movementanalysator.cpp
tomburns: **Severity: minor — should not gate the merge**, and quite possibly void.

I'm leaving this one open, unlike the other #83 threads, only because it contains a question that changes the outcome: if Excelsior is N=18 and kangaroo N=36 as this PR's design doc states, my original finding was simply wrong and there is nothing to restate. Resolving it now would bury that question rather than settle it.

A one-line confirmation of the two piece counts closes it either way.

tomburns (approved):

PR #86, opened 2026-09-19
ci: gate on an x86-64 Clang build and a ThreadSanitizer run

tomburns:
Adds two CI jobs on top of the stack (#78 → #79 → #81 → #82 → #83 → #84), covering two defect classes the current jobs are structurally unable to see.

**Both jobs fail on this stack as it stands. That is the intended signal, not a bug in the jobs** — each one reproduces a confirmed defect. See "What these jobs currently catch" below.

## `clang-x86-64`

Builds with Clang on x86-64. Every existing job compiles the x86 intrinsic code with GCC or not at all: Linux and Windows are GCC, and `macos-latest` is arm64, where those blocks are excluded by the preprocessor. So no job compiles them with Clang.

This matters because `#pragma GCC target(...)` is a GCC extension. Clang parses it as an unknown pragma, ignores it, and then every intrinsic in the block fails because the target feature was never enabled.

Verified against these sources with Apple Clang 21, `-arch x86_64`:

| file | on #82 | on #83 |
|---|---|---|
| `src/lib/simd_exact_cover.cpp` | 6 errors | 6 errors |
| `src/lib/simd_huang_cover.cpp` | 6 errors | 6 errors |
| `test/test_simd_exact_cover.cpp` | 1 error | 1 error |
| `src/lib/movementanalysator.cpp` | 0 | **20 errors** |

```
error: always_inline function '_mm256_load_si256' requires target feature 'avx',
  but would be inlined into function 'filterRowsAvx2' that is compiled without
  support for 'avx'
error: AVX vector return of type '__m256i' without 'avx' enabled changes the ABI
```

`--werror` is deliberate: meson applies it to this project only, not to subprojects, and the ignored-pragma warnings are themselves the earliest signal.

The fix is mechanical and verified — `__attribute__((target("avx2")))` per function compiles clean for x86-64 and emits `vptest`. The `push_options`/`pop_options` wrappers can be deleted.

## `tsan-parallel`

Runs the `[parallel]` tag under ThreadSanitizer. No job currently runs a sanitizer. `timeout-minutes` is set so a lost wakeup in the solver or disassembler pool fails the job rather than hanging the runner.

## Why a TSan job alone would not have worked

Running TSan against the existing parallel tests reports **zero races**, and that is a false negative. Instrumenting the guard shows why:

```
[PROBE a0] reached_callback=12   avoidTransformed_true=0
[PROBE a1] reached_callback=100  avoidTransformed_true=0
```

`avoidTransformedAssemblies` is only enabled by `checkForTransformedAssemblies()`, called from `createMatrix()` and only when the result shape has a symmetry breaker. Neither PelikanBurr nor CubeInCage qualifies, so the existing tests **never execute the racy branch at all**.

So this PR also adds a test case on `Bermuda.xmpuzzle`, which does enable symmetry breaking, and asserts that premise rather than trusting it: the `keepRotations = true` run must yield strictly more assemblies, otherwise symmetry breaking is off and the case has silently stopped covering the path.

Two further details, both documented inline because both silently disable the test:

- **The parallel run reloads the puzzle.** The `mutable` caches in `voxel_c` are first-touch. A parallel run performed after a serial run on the same puzzle object races on already-warm caches and reports nothing. I hit exactly this while writing the test — the first version was green because its own serial baseline had warmed the caches.
- **Assemblies are compared as multisets, not counts.** A run that loses one assembly and duplicates another passes a count check.

## What these jobs currently catch

**`tsan-parallel`** — deterministic, 6/6 consecutive runs on this branch:

```
JOB EXIT = 134   races = 1
  1x voxel.cpp:237 in voxel_c::getHotspot(unsigned char, int*, int*, int*) const
```

At the #79/#81 point in the stack, where the worker calls `smallerRotationExists()` directly, the same puzzle produces **45** races across six distinct lines in `getHotspot`/`getBoundingBox`, all arriving via `assembly_c::transform` from two worker threads. The unlocked call is still present at the tip, now consolidated into `handleSolution()` (`assembler_0.cpp:1238`).

**`clang-x86-64`** — fails on the `simd_exact_cover` / `simd_huang_cover` / `movementanalysator` errors above.

Both go green once the `#pragma GCC target` → `__attribute__((target))` change and the shared voxel cache pre-warm land. Inline review comments on #79, #81, #82 and #83 carry the details.

## Verification

- `meson test --suite fast --suite slow` — all green (3/3) in a normal build
- New test passes in a normal build in 0.26 s
- TSan detection confirmed deterministic over 6 consecutive runs
- Workflow YAML parses; job order is `build-linux, clang-x86-64, tsan-parallel, coverage, build-windows, build-macos, release`

`AGENTS.md` gains a "Continuous Integration Gates" section recording the intrinsics rule and the requirement to keep a symmetry-breaking puzzle in the `[parallel]` tag, since that constraint is invisible from the test alone.


akoehn:
Rebased on top of `huang-simd-tiers`. The new Bermuda test case exercising parallel assembly on a symmetry-breaking puzzle (`[assembler][parallel][tsan]`) passes cleanly, confirming that the `voxel_c` cache pre-warming and multiset synchronization fixes resolve the race.

akoehn (dismissed):

PR #87, opened 2026-09-19
feat(solver): extend SimdHuangCover to 32768 columns and enable hole limits (1.8x–23.5x speedup)

akoehn:
## Summary

Extends **`SimdHuangCover`** (Assembler 1 / duplicate shapes, piece ranges, variable voxels) from a hardcoded 256-column limit to templated tiers up to **32,768 columns**, lifts the conservative hole-limit restriction in `canUseSimd()`, and introduces major algorithmic optimizations to the search loop.

This unlocks hardware-vectorized SIMD exact cover (AVX-512, AVX2, NEON) for all Assembler 1 puzzles with $> 256$ columns, yielding **1.8x to 23.5x speedups** over classic DLX.

---

## Benchmark Results: SIMD vs Classic DLX Baseline

Measured across Assembler 1 puzzles with $> 256$ columns (verified 100% assembly parity bit-for-bit):

| Puzzle | Matrix Columns | SIMD Tier | DLX Baseline (Single / Multi) | Optimized SIMD (Single / Multi) | Single-Core Speedup | Multi-Core Speedup | Assemblies Found | Iteration Reduction |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| **`Simplicity.xmpuzzle`** | 356 | `SimdHuangCover512` | 1.970s / 0.701s | **0.084s / 0.066s** | **23.49x** | **10.59x** | 188 | 5.1M → 19.4k |
| **`HexSticks.xmpuzzle`** | 2,259 | `SimdHuangCover4096` | 0.842s / 0.316s | **0.063s / 0.060s** | **13.41x** | **5.25x** | 33 | 1.8M → 5.7k |
| **`Tippy.xmpuzzle`** | 295 | `SimdHuangCover512` | 0.148s | **0.027s** | **5.26x** | — | 460 | 274k → 6.5k |
| **`BottomLine.xmpuzzle`** | 355 | `SimdHuangCover512` | 0.064s | **0.011s** | **4.05x** | — | 76 | 157k → 2.6k |
| **`Third_Times_the_Charm.xmpuzzle`** | 521 | `SimdHuangCover1024` | 4.157s / 2.315s | **2.292s / 0.976s** | **1.81x** | **2.37x** (<1s) | 71 | 440k → 19.6k |

---

### 1. Templated Tiers up to 32,768 Columns
- Introduced `ISimdHuangCover` abstract base class with virtual `solve()`, `solveSubtree()`, and `parallelSolve()`.
- Templated `SimdHuangCover<BitsetType>` on `BitsetType` (`SimdBitset256` through `SimdBitset32768`).
- Added AVX-512 `filterRowsAvx512` path using `_mm512_test_epi64_mask`.
- Explicit template instantiations for all 8 tiers (256, 512, 1024, 2048, 4096, 8192, 16384, 32768).
- Updated `assembler_1_c::canUseSimd()` limit from 256 to 32,768 columns and `createSimdSolver()` with dynamic tier selection ladder.

### 2. Enabled Hole Limits in `canUseSimd()`
- Removed the conservative restriction `if (holes < holeColumns.size()) return false;` in `canUseSimd()`.
- `SimdHuangCover` already implements hole limit pruning in `search()` (`if (empty_holes > holes) return;`), allowing all puzzles with variable voxels (`+`) and hole limits to safely use SIMD.

### 3. Search Loop Optimizations
- **$O(1)$ Candidate Row Matching:** Replaced the linear `for (c : cand.columns)` check with an $O(1)$ bit test (`cand.voxel_mask.test(best_col - 1)`) or shape equality test (`cand.shape_col == best_col`).
- **Eliminated Redundant Disjointness Checks:** Removed `is_disjoint_scalar()` in the candidate loop since `filterRows()` already guarantees disjointness.
- **Depth-Gated Goal Check:** Gated the expensive `containsAll()` and column scan behind `ctx.current_solution.size() >= total_min_pieces`, skipping it on >95% of non-leaf search nodes.
- **Bitset-Accelerated Unplaced Voxel Scanning:** Merged dead-end pruning and MRV column selection into a single pass. Unplaced voxels are scanned directly from `required_voxels & ~ctx.placed_voxels` using `std::countr_zero`, completely skipping placed and non-required voxels.

### 4. Documentation & Analysis
- **`bench/puzzle_properties.md`:** Comprehensive catalog analyzing all 289 problems across `examples/` and `puzzles/BTFiles/`, categorizing by assembler backend, piece constraints, column counts, and SIMD tier.
- **`design/2026-09-19-assembler-1-simd-incremental-design.md`:** Detailed architectural evaluation of sparse DLX vs dense SIMD.

---

## Verification

- `just test-all` passes (3/3: fast Catch2, Python bindings, stress cases).
- `just check` passes (cppcheck clean, 0 warnings).
- Unit tests added for `SimdHuangCover512`, `1024`, `4096`, and `32768` testing both `solve()` and `parallelSolve()`.
- Stacked on `disassembler-pool` (#84).

---

## Known behaviour change: resumability

Widening `canUseSimd()` moves a large new class of assembler-1 puzzles onto the SIMD path, and that path is **not resumable**. Stopping a solve on a newly-eligible puzzle and saving the `.xmpuzzle` produces a position that `setPosition()` refuses with `ERR_CAN_NOT_RESTORE_INTERRUPTED`.

Refusing is the correct conservative behaviour — the alternative is replaying the search and duplicating every assembly — but it is a loss relative to the DLX path these puzzles used to take. `examples/CubeInCage.xmpuzzle` is a concrete case, which is why `test/test_solver.cpp:927` is pinned to `BURRTOOLS_NO_SIMD=1`.

Tracked as #91. Per-tier memory footprint (`Row::voxel_mask` at the top tier, and `scratch_active_rows` being sized from the column count while indexed by search depth) is tracked as #92.

## Review follow-ups addressed in 028e124b

- **AVX-512 ignored the SIMD kill switches.** `BURRTOOLS_NO_SIMD` / `BURRTOOLS_NO_AVX2` did not clear `use_avx512`, so on an AVX-512 host `BURRTOOLS_NO_AVX2=1` selected *wider* SIMD instead of falling back, leaving the scalar kernel unreachable from tier 512 up. Now mirrors `SimdExactCover`.
- **`ISimdHuangCover::activeKernel()`** exposes the dispatch decision, with a `TEMPLATE_TEST_CASE` over five tiers locking all three kill switches — every kernel returns identical solutions, so this cannot be caught by comparing results.
- **The multi-threaded SIMD completion path** now clears `next_row_stack` / `task_stack` like the DLX path, so a finished run no longer looks unstarted to `save()`.
- **`setColumnBounds()` is idempotent**: a repeated call replaces its `total_min_pieces` contribution rather than double-counting it (which would have silently suppressed every solution), and no longer lists the same hole column twice.
- **The narrowed voxel-column contract** (`min_weight <= 1`, since the goal check is a presence test) is documented and asserted.


tomburns (commented):
Re-posting my earlier review as inline comments, with suggested fixes where the change is mechanical.

Reviewed `gh pr diff 87` against merge base `fab68672` — 9 files, with the substantive changes in `src/lib/simd_huang_cover.{h,cpp}`, `src/lib/assembler_1.{h,cpp}` and the tests. **1 high, 3 medium, 3 low.** The speedup itself looks real and the core algorithm changes hold up; the findings are about the blast radius of widening eligibility.

### Things I checked and believe are correct

- **Removing `is_disjoint_scalar(placed_voxels, cand.voxel_mask)`** from `search()`/`generateTasks`: the invariant holds inductively — depth 0 has an empty `placed_voxels`, every deeper candidate set comes from `filterRows` which enforces disjointness against the newly placed row, and `solveSubtree` applies its prefix through the same `filterRows`.
- **The `covers_best` rewrite**: `best_col` can only be a shape column, the range column, or a required voxel, so the three-way ternary is exhaustive; hole columns are excluded from `required_voxels` because their `min_weight` is 0.
- **Enabling hole limits**: with `canUseSimd()`'s surviving `min == max` per-part requirement, total piece volume is fixed, so every assembly has exactly `holes = res_filled - min` holes and the existing `hole_columns` pruning is sound — the goal check never needs to re-count them.
- **Alignment for `_mm512_load_si512`**: `SimdBitset` is `alignas(64)`, `voxel_mask` is `Row`'s first member, and the build is `cpp_std=c++20` (`meson.build:9`), so `std::vector<Row>` uses aligned allocation. `std::countr_zero` is covered by the new `<bit>` include.

---

src/lib/simd_huang_cover.cpp
```diff
@@ -25,22 +25,28 @@
...
+  bt_assert(num_cols <= BitsetType::NUM_WORDS * 64);
   columns.resize(num_columns + 1);
 
 #if (defined(__x86_64__) || defined(_M_X64)) && (defined(__GNUC__) || defined(__clang__))
   if (!std::getenv("BURRTOOLS_NO_SIMD") && !std::getenv("BURRTOOLS_NO_AVX2")) {
     use_avx2 = __builtin_cpu_supports("avx2");
   }
+  if (!std::getenv("BURRTOOLS_NO_AVX512")) {
+    use_avx512 = __builtin_cpu_supports("avx512f");
+  }
```
tomburns, line 45: ### 🔴 High — AVX-512 detection ignores the SIMD kill switches

The `NO_AVX512` check stands alone: neither `BURRTOOLS_NO_SIMD` nor `BURRTOOLS_NO_AVX2` gates it. `SimdExactCover` (`src/lib/simd_exact_cover.cpp:39-45`) clears **both** flags for those two variables; this copy doesn't.

Concrete effect: on an AVX-512 host `filterRows` dispatches to `filterRowsAvx512` first (`simd_huang_cover.cpp:289`), so `BURRTOOLS_NO_AVX2=1` — the documented way to A/B against the scalar kernel — silently runs *wider* SIMD instead of narrower. The scalar/NEON fallback inside `filterRows` becomes unreachable, and therefore untested, for every tier from 512 up.

Nothing currently catches this: the unit tests in `test_simd_exact_cover.cpp` construct these solvers directly with no env set, and the corpus A/B test still compares two real paths only because `assembler_1_c::canUseSimd()` short-circuits on `BURRTOOLS_NO_SIMD` before a solver is ever constructed.

```suggestion
  if (!std::getenv("BURRTOOLS_NO_SIMD") && !std::getenv("BURRTOOLS_NO_AVX2")) {
    use_avx2 = __builtin_cpu_supports("avx2");
    use_avx512 = __builtin_cpu_supports("avx512f");
  }
  if (std::getenv("BURRTOOLS_NO_AVX512")) {
    use_avx512 = false;
  }
```

src/lib/assembler_1.cpp
```diff
@@ -2721,12 +2724,9 @@ bool assembler_1_c::canUseSimd(void) const {
   if (debug)
     return false;
 
-  if (headerNodes - 1 > 256)
+  if (headerNodes - 1 > 32768)
     return false;
```
tomburns, line 2735: ### 🟠 Medium — widened eligibility silently costs resumability

Dropping the `holes < holeColumns.size()` bail-out and raising the cap from 256 to 32768 moves a large new class of assembler-1 puzzles onto the SIMD path — and that path is not resumable: `simdSearch()` sets `parallelInterrupted = abbort` (line 2845), so `setPosition()` then returns `ERR_CAN_NOT_RESTORE_INTERRUPTED`.

The user-visible consequence is that stopping a solve on any newly-eligible puzzle and saving the `.xmpuzzle` now produces a position the reload refuses. That's a behaviour change this PR doesn't mention.

Two ways forward, both fine by me — but worth picking deliberately:
1. Make a SIMD position saveable (mirroring what the DLX path does), or
2. Document the loss of resume as a known tradeoff of enabling the wider tiers.

See also the stale comment at `test/test_solver.cpp:1038` — "holes, variable voxels and >512 columns all disqualify it" is no longer true after this change.

test/test_solver.cpp
```diff
@@ -892,6 +924,7 @@ TEST_CASE("Parallel assembler 1: an interrupted search is not restored as resuma
  */
 TEST_CASE("Parallel assembler 1: the interrupted flag does not break normal restore",
           "[assembler][parallel][resume]") {
+  ScopedEnv env("BURRTOOLS_NO_SIMD", "1");
```
tomburns, line 927: ### 🟠 Medium (same issue as `assembler_1.cpp:2727`) — this line is the evidence, not the fix

This test runs `examples/CubeInCage.xmpuzzle` single-threaded and previously exercised DLX: that puzzle has 39 variable voxels against ~70 units of piece volume and 99 result cells, i.e. `holes == 29 < 39`, which the removed guard rejected. Widening eligibility moves it onto the SIMD path, where the position no longer round-trips — hence the new `ScopedEnv`.

Pinning the test to `BURRTOOLS_NO_SIMD=1` keeps it green but stops it from covering the configuration real users now get. If resume genuinely can't survive on the SIMD path yet, I'd suggest keeping this case *and* adding a companion that asserts the new SIMD behaviour explicitly, so the tradeoff is encoded rather than hidden.

src/lib/assembler_1.cpp
```diff
@@ -2758,7 +2758,24 @@ std::unique_ptr<SimdHuangCover256> assembler_1_c::createSimdSolver(void) const {
...
+    solver = std::make_unique<SimdHuangCover2048>(num_cols, num_shapes);
+  } else if (num_cols <= 4096) {
+    solver = std::make_unique<SimdHuangCover4096>(num_cols, num_shapes);
+  } else if (num_cols <= 8192) {
+    solver = std::make_unique<SimdHuangCover8192>(num_cols, num_shapes);
+  } else if (num_cols <= 16384) {
+    solver = std::make_unique<SimdHuangCover16384>(num_cols, num_shapes);
+  } else {
+    solver = std::make_unique<SimdHuangCover32768>(num_cols, num_shapes);
+  }
```
tomburns, line 2785: ### 🟠 Medium — per-row and per-task memory scales with the tier, not the problem

The ladder selects on column count, but the memory cost rides on the tier. `SimdHuangCover<SimdBitset32768>::Row` embeds a 4096-byte `voxel_mask`, so a puzzle landing in the top tier costs ~4 KB **per placement row**; assembler-1 matrices routinely have 10^5 rows, i.e. hundreds of MB allocated before the search even starts — where the old code simply fell back to DLX.

`SubtreeTask` compounds it: it copies a whole `SearchContext`, whose `scratch_active_rows` is `num_columns + 16` `std::vector`s. At 32768 columns that's ~790 KB of empty vector headers **per task**, and `generateTasks` can emit one task per child of the root (thousands). At 256 columns this was ~6.5 KB/task and invisible; at the new tiers it's a plausible OOM on a machine that previously just ran slower.

Two mitigations worth considering:
- Size `scratch_active_rows` from `num_columns` rather than the tier's `NUM_WORDS * 64`, so a 600-column puzzle in the 1024 tier doesn't pay 1024 columns' worth of headers per task.
- Cap eligibility in `canUseSimd()` on estimated footprint, not just column count, e.g.

```cpp
  /* A 32768-tier Row carries a 4 KB voxel_mask; refuse the SIMD path rather
   * than allocating hundreds of MB where DLX would have coped. */
  if (static_cast<uint64_t>(estimatedRows) * tierRowSize(num_cols) > kSimdMemoryBudget)
    return false;
```

src/lib/assembler_1.cpp
```diff
@@ -2605,6 +2605,9 @@ void assembler_1_c::parallelMultiSearch(unsigned int workers) {
       parallelTasks.clear();
       taskCompleted.clear();
       emittedSignatures.clear();
+      parallelInterrupted = false;
```
tomburns, line 2615: ### 🟡 Low — SIMD completion path doesn't clear the stacks

The DLX branch immediately below (lines 2698-2704) also does `next_row_stack.clear(); task_stack.clear();` on the success path; this one doesn't. The asymmetry looks unintentional.

After a *completed* multi-threaded SIMD run the object therefore still looks like an unstarted search, and `save()` writes `interrupted = 0` next to the root position. Reloading that, `setPosition()` accepts it and `assemble()` replays the whole search, re-reporting every assembly — exactly the duplication `parallelInterrupted` exists to prevent. Narrow in practice, since `problem_c` only saves while `SS_SOLVING`, but cheap to close.

```suggestion
      next_row_stack.clear();
      task_stack.clear();
      parallelTasks.clear();
      taskCompleted.clear();
      emittedSignatures.clear();
      parallelInterrupted = false;
```

src/lib/simd_huang_cover.cpp
```diff
@@ -62,20 +68,25 @@ void SimdHuangCover256::setColumnBounds(
   columns[col].is_range = is_range;
   columns[col].is_hole = is_hole;
 
+  if (is_shape) {
+    total_min_pieces += min_w;
+  }
```
tomburns, line 95: ### 🟡 Low — `total_min_pieces` accumulates unconditionally

This runs on every call, so calling `setColumnBounds()` twice for the same shape column double-counts. The gate at line 683 would then silently suppress *every* solution rather than failing loudly — a nasty way to lose an hour.

Unreachable from `assembler_1_c` today, which sets each column exactly once, but `SimdHuangCover` is now a templated public component behind `ISimdHuangCover`, so the precondition is worth stating at minimum:

```suggestion
  if (is_shape) {
    /* Accumulates: setColumnBounds() must be called at most once per column.
     * A second call for the same shape double-counts here, and the goal-check
     * gate in search() would then reject every candidate solution. */
    total_min_pieces += min_w;
  }
```

Recomputing `total_min_pieces` from `columns[]` at solve time would make it a non-issue instead.

src/lib/simd_huang_cover.cpp
```diff
@@ -595,25 +680,29 @@ void SimdHuangCover256::search(
...
-
-  if (!ctx.placed_voxels.containsAll(required_voxels)) {
-    all_fulfilled = false;
-  } else {
-    for (unsigned int c : active_column_list) {
-      if (ctx.col_weights[c] < columns[c].min_weight || ctx.col_weights[c] > columns[c].max_weight) {
-        all_fulfilled = false;
-        break;
+  if (ctx.current_solution.size() >= total_min_pieces) {
+    if (ctx.placed_voxels.containsAll(required_voxels)) {
```
tomburns, line 711: ### 🟡 Low — goal check validates voxel columns at "weight ≥ 1"

Dropping the `active_column_list` sweep in favour of shape columns + range + `containsAll` is a good speedup, but it changes what's checked for voxel columns: `placed_voxels.containsAll(required_voxels)` is a presence test, so a voxel column with `min_weight > 1` would be reported satisfied at weight 1.

Not reachable from `assembler_1_c`, where voxel columns are always min/max 1 — flagging it only because this class is now a public templated component, so the narrowed contract should be explicit (a comment, or a `bt_assert` in `setColumnBounds` that voxel columns have `min_weight <= 1`).

src/lib/simd_huang_cover.cpp
tomburns: Fixed in 028e124b. Gating now mirrors `SimdExactCover`: `BURRTOOLS_NO_SIMD` and `BURRTOOLS_NO_AVX2` clear both flags, `BURRTOOLS_NO_AVX512` clears only the wider one.

On the "nothing currently catches this" half — no result comparison can, since every kernel returns the same solutions. So I exposed the dispatch decision itself as `ISimdHuangCover::activeKernel()` and added a `TEMPLATE_TEST_CASE` over five tiers asserting all three kill switches.

Mutation-checked that the test has teeth: breaking the NEON gating the same shape the AVX-512 bug had (`use_neon = !getenv("BURRTOOLS_NO_NEON")`) fails it with `"neon" == "scalar"`. Caveat — I am on arm64, so the AVX sections compile out here and only bite on x86 CI; the `NO_SIMD` section is the one with local coverage.

src/lib/assembler_1.cpp
tomburns: Confirmed and filed as #91 rather than fixed here — making a SIMD position saveable means serializing `SimdHuangCover` search state, which is its own PR.

The stale comment is fixed in 028e124b: `test/test_solver.cpp:1038` now reads "a range column, >32768 matrix columns, or variable voxels combined with a shape whose min differs from its max", which is what `canUseSimd()` actually rejects after this change.

test/test_solver.cpp
tomburns: Agreed on the diagnosis — the `ScopedEnv` here is the symptom. Tracked in #91 along with the resume work itself, including removing this pin once SIMD positions round-trip and adding the `assembler_1_c` companion to the `assembler_0_c` case at `:1046`.

src/lib/assembler_1.cpp
tomburns: Filed as #92. One correction on the second mitigation, though — `scratch_active_rows` is already sized from `num_columns`, not from the tier: `simd_huang_cover.cpp:334`, `:364` and `:448` all do `resize(num_columns + 16)`. A 600-column puzzle in the 1024 tier pays 616 headers, not 1024, so that sizing is not tier-driven.

The real waste is adjacent: the vector is indexed by **search depth**, and each recursion places exactly one row, so depth is bounded by the total placeable pieces (sum of `max_weight` over shape columns), not by the column count. At 32768 columns that is ~790 KB of empty vector headers per `SearchContext` where a few hundred bytes would do — so your ~790 KB/task figure is right, just for a different reason, and the fix is a tighter bound rather than a different source.

The `Row::voxel_mask` half stands exactly as written. Both are in #92, along with the constraint that the allocation has to stay up-front (the comment at `:793` — `curr_active` is a reference into it).

src/lib/assembler_1.cpp
tomburns: Fixed in 028e124b — the SIMD completion path now clears `next_row_stack` and `task_stack` too. The asymmetry was unintentional.

src/lib/simd_huang_cover.cpp
tomburns: Fixed in 028e124b, taking the "non-issue" route rather than the comment: `setColumnBounds()` now captures the column`s existing contribution before overwriting and subtracts it, so a repeat call replaces rather than accumulates. O(1), so the hot-path read at `search()` is unchanged.

While there: `hole_columns.push_back(col)` had the same shape of problem — a repeated call would list the same hole column twice and inflate the empty-hole count that `search()` prunes on. Guarded it the same way.

src/lib/simd_huang_cover.cpp
tomburns: Fixed in 028e124b with both, since the assert alone does not explain itself: a comment stating that the goal check is a presence test and therefore cannot distinguish weight 1 from weight n, plus `bt_assert(!is_voxel || min_w <= 1)` in `setColumnBounds()`. Suite is green, so nothing in the corpus violates it.

tomburns (approved):

PR #94, opened 2026-09-20
burrTxt2: fix the hang, the abort on grids without a disassembler, and assembler 1 reporting no assemblies

tomburns:
Three ways BurrTools gave a wrong or useless answer with no diagnostic. All three reproduce on shipped examples.

## 1. burrTxt2 hangs forever on a file that is already solved

`burrTxt2 <file>` — no `-R` — wedged with no output and no CPU use on any file whose problem is stored finished. Every example in `examples/` is stored that way, so `burrTxt2 examples/PelikanBurr.xmpuzzle` reproduces it.

1. Without `-R`, burrTxt2 continues the problem, so `solveThread_c::run()` builds an assembler and calls `problem_c::setAssembler()`.
2. `setAssembler` asserts `solveState == SS_UNSOLVED` for a problem with no saved assembler state. One loaded in `SS_SOLVED` fails that, and `bt_assert` throws.
3. `run()` catches it and parks the worker in `ACT_ASSERT`, then the thread exits.
4. burrTxt2's poll loop only left on `ACT_FINISHED` or `ACT_ERROR`. The worker was gone, but main kept polling — silently, because the progress `switch` has no case for the states a run actually ends in.

Release builds define `NDEBUG`, which compiles `bt_assert` to a no-op, so step 2 never fires there. It shows up in a debug build.

**Fix**

- `solveThread_c::stopped()` now reports `ACT_ASSERT`. An assert ends the worker as surely as the other terminal states, and a caller waiting for the thread to finish has to be able to see that.
- burrTxt2 polls with `stopped()` and reports every state a run can end in. Reporting moved after the loop, where it is reachable — the old in-loop `ACT_ERROR` handling was dead code, excluded by the loop condition directly above it. A failed run now exits non-zero.
- `problem_c::canStartSolving()` lets callers honour `setAssembler`'s precondition. burrTxt2 checks it and reports what the file holds instead of starting a run that cannot work. This mirrors the GUI, which keeps *Continue* disabled unless the problem is resumable — which is why the GUI never hit this.

## 2. burrTxt2 aborts on `-d` for a grid that has no disassembler

`burrTxt2 -d examples/BrokenSticks.xmpuzzle` died with `libc++abi: terminating due to uncaught exception of type assert_exception: cache`.

Rhombic, sphere and tetra-octa grids have no movement cache — `gridType_c::getMovementCache()` returns `0` — and `movementAnalysator_c`'s constructor asserts on that. The disassembler is built by the `solveThread_c` **constructor**, which runs on the main thread, outside `run()`'s try/catch, so the assert escaped `main` and aborted the process. The GUI checks `CAP_DISASSEMBLE` before starting; burrTxt2 checked nothing.

**Fix**

- burrTxt2 checks `CAP_ASSEMBLE`, and `CAP_DISASSEMBLE` when `-d` was given, right after loading the file, and names the missing one. Assembling those puzzles always worked and still does.
- A `-b` index the file has no problem for is reported instead of asserting its way out of `puzzle_c::getProblem()`.
- `main()` catches `assert_exception` around everything else, so an assert from outside the solver thread prints where it came from rather than aborting. That is the fallback, not the fix — the checks above are.

## 3. Assembler 1 reports no assemblies on a reduced matrix

`examples/12PieceSeparation` and `examples/AlPackino` came back with 0 assemblies instead of 1. The DLX fallback (`BURRTOOLS_NO_SIMD=1`) was always correct, so the answer depended on which search ran.

`reduce()` ends in `clumpify()`, which drops every column covered by exactly the same rows with the same weights as an earlier one; `remove_column()` unlinks those columns' nodes from the rows as well. `createSimdSolver()` registered columns `1..num_cols` regardless, so the dropped ones arrived as columns no row could ever cover — 229 of 259 for one six-piece board burr. A *required voxel* among them made `expand()` read the puzzle as unsatisfiable at the root, emit no subtree tasks at all, and finish having found nothing; `search()` never even ran.

Bisected to 4799306c (#87), which raised the SIMD column cap from 256 to 32768. The mismatch was always there — that commit started routing these matrices through it.

**Fix**

- `createSimdSolver()` registers only the columns still linked into the header ring. Those are the ones the rows can still reach, and the ones `assembler_1_c::rec()` itself walks.
- Confirming the hole handling #87 also enabled turned up a second disagreement with DLX: `SimdHuangCover` applied the hole budget when pruning the search below a node, but its goal test ran first and never consulted the budget, so it returned assemblies with more holes than the puzzle allows. It now counts the unfilled hole columns before reporting a solution, as `assembler_1_c::rec()` does.

## Tests

- `test/cli_smoke.py`, run by meson as `burrtxt2_cli_smoke`: drives the real binary, because issues 1 and 2 live in `main()` — which runs are refused, and that every run terminates and says why. Each run has a timeout, so a wedged binary fails the test instead of hanging the suite.
- A Catch2 case for the library half of issue 1: a `solveThread_c` started on an already-solved problem must reach `stopped()`.
- A `SimdHuangCover` hole-budget case. Nothing called `setHoles()` anywhere before.
- An assembler-level case pinning the SIMD search to the DLX result on a **reduced** matrix. The existing `solvePuzzle()` helper never calls `reduce()`, which is why none of issue 3 was covered.

Every one of these was confirmed red before its fix and green after. Full suite passes: 5/5 meson test targets.


akoehn (dismissed):

akoehn:
I see some comments that describe what the code did wrong here in the past. I do not think that kind of comment will be helpful in a year or so

tomburns:
Agreed, and rewritten in fd1de6a2f. Seven comments across `burrTxt2.cpp`, `assembler_1.cpp`, `cli_smoke.py`, `test_solver.cpp` and `test_simd_exact_cover.cpp` were narrating the broken version rather than the current one.

They now state what has to hold and why a future edit cannot drop it. The clearest case:

```c
-   * Registering them here anyway left the solver with columns no row can ever
-   * cover, so a required voxel among them looked unsatisfiable and the search
-   * gave up at the root and reported no assemblies at all.
+   * unlinks those columns' nodes from the rows along with it. The solver has to
+   * see the same matrix the rows describe: a column no row can reach reads as
+   * an unsatisfiable constraint and ends the search at the root.
```

Same treatment for the hole budget in `simd_huang_cover.cpp` (it applies to the assembly being reported, not only to the search below the node) and the post-loop report in `burrTxt2.cpp` (every state that means the worker has left `run()` needs an answer there).

The incident stays in the commit messages, which is where it is still worth having — a bisect lands on the commit, not the comment.

Two comments I left alone, since they describe present mechanism rather than past behaviour: the `CAP_DISASSEMBLE` check in `burrTxt2.cpp` explains why the check has to happen before the solver thread is constructed (the assert fires on the main thread, where nothing catches it), and `stopped()` in `solvethread.h` explains why `ACT_ASSERT` belongs in the set. Say the word if either still reads as history and I will trim them too.

Full suite green, 5/5.

PR #96, opened 2026-09-21
fix(gui): default the export path to the puzzle's folder, and report write failures

tomburns:
## The problem

The image export dialog's path field starts empty, so the output filename is relative to the working directory. That is fine when BurrTools is started from a shell, and useless when it is started the way macOS users start applications: a bundle launched from Finder or the Dock runs with a working directory of `/`, which is read-only.

Verified rather than assumed — `lsof` on a running bundled BurrTools:

```
COMMAND     PID   USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
burrtools 83292 tburns  cwd    DIR   1,19      704    2 /
```

The STL dialog has the same problem one step removed: it defaults to `"."`, which is the same working directory.

Worse, `imageexport.cpp` discarded `saveToPNG()`'s return value. A failed write only printed to `stderr` — which a bundled application does not have — so the dialog counted through "save page 1, 2, 3…", reported **Done**, and left the user with no files and no explanation.

## The change

Both dialogs now default their path to the directory the current puzzle was loaded from, falling back to the home directory when the puzzle has never been saved.

A failed write now aborts the export and reports it, naming the path. Once, not once per page: if the first page cannot be written, none of them can, and a dialog per page would be an ordeal.

## On the tests

The directory logic is a pure function in `src/tools/` rather than inline in the dialogs, because `src/tools` is linked into `test_burrtools` while `src/gui` is not. It is the first piece of this area that carries real unit tests.

They cover what the old inline code got wrong:

- a path that is already a directory (`/home/tom/puzzles/`) — its own answer, not its parent's
- a bare filename — empty result, so the caller can tell the difference and supply its own fallback
- a root-level file (`/burr.xmpuzzle` → `/`, not `""`)
- `.` surviving as a relative directory rather than being read as an extension separator
- **Windows separators** — the cross-compiled Windows build gets backslash paths from `Fl_Native_File_Chooser`, and a `/`-only split would read every one of them as a bare filename

Choosing the fallback stays in the dialogs (they use `homedir()`), so the helper stays pure and testable.

## Verification

`just build && just test-all && just check && just build-werror` all clean; 4/4 suites, 10 new assertions across 5 cases.

Not verified: the dialogs themselves. This sandbox has no Accessibility or Screen Recording permission, so the failure alert and the pre-filled path field have not been seen on screen — worth a look before merging.


src/gui/imageexport.cpp
```diff
@@ -263,7 +267,16 @@ void imageExport_c::nextImage(bool finish) {
     else
       snprintf(name, 1000, "%s%s%03u.png", Pname->value(), Fname->value(), curPage);
 
-    i->saveToPNG(name);
+    /* saveToPNG reports failure by returning 0 and printing to stderr.
+     * That return used to be discarded, so an export that could not write
```
akoehn: same as always: do not document the past but the present

src/gui/imageexport.cpp
```diff
@@ -579,7 +614,26 @@ void imageExport_c::cb_SzUpdate(void) {
...
 }
 
-imageExport_c::imageExport_c(puzzle_c * p) : LFl_Double_Window(false), puzzle(p), working(false), state(0), i(nullptr) {
+imageExport_c::imageExport_c(puzzle_c * p, const std::string & puzzleFile) : LFl_Double_Window(false), puzzle(p), exportDir(directoryOfFile(puzzleFile)), working(false), state(0), i(nullptr) {
+
+  /* Where the images go when the user does not say. The puzzle's own folder
+   * is the answer they almost always want, and for an unsaved puzzle the
+   * home directory at least exists and is writable.
+   *
+   * The field used to start empty, which made the output name relative to
```
akoehn: same, document what IS.

src/tools/filepath.cpp
```diff
@@ -0,0 +1,65 @@
...
+   * path examined with only '/' in mind looks exactly like a bare filename,
+   * which would send the export dialogs to the home directory instead of
+   * the folder the puzzle came from.
+   */
+  bool isSeparator(char c) {
+    return c == '/' || c == '\\';
+  }
+}
+
+std::string directoryOfFile(const std::string & path) {
```
akoehn: There is no standard library method for this? C++ is weird

akoehn (approved):
approved but tackle comments first

src/gui/imageexport.cpp
tomburns: Rewritten to state what the code does. The stderr point stays because it is why the return value is checked at all, not history:

    /* saveToPNG reports failure by returning 0 and printing to stderr. The
     * return value is what matters here: a program started from a macOS
     * bundle or a Windows shortcut has no stderr anyone can read, so the
     * path is recorded for PostDraw() to report in a dialog instead.
     */

src/gui/imageexport.cpp
tomburns: Done. The three paragraphs of how-it-used-to-be are gone; what is left describes the two branches and why the home directory is the fallback:

    /* The path field starts at the folder the puzzle was loaded from, which is
     * where the user almost always wants the images. An unsaved puzzle has no
     * folder, so the home directory stands in: it exists and is writable,
     * neither of which holds for the working directory -- an application
     * launched from a macOS bundle or a Windows shortcut is given "/".
     *
     * homedir() ends in a separator, so it goes through parent_path() as well:
     * that drops the separator and leaves both branches in the same shape, so
     * the name assembly in nextImage() cannot produce a doubled one.
     */

`stlexport.cpp` had the same defect in its constructor comment one file over. Fixed there too rather than leaving it for the next review.

src/tools/filepath.cpp
tomburns: There is one, and you were right to ask: `std::filesystem::path::parent_path()`. This codebase already depends on `<filesystem>` — `src/tools/fileexists.cpp` right next door, and `puzzle_c::load`/`save` — so there was never even a dependency argument for hand-rolling it.

I checked `parent_path()` against all five test cases the file carried. It matches every one of them: trailing separator, bare filename, root-level file, `./x`, empty input.

The single divergence is the case the comment used to justify the whole function — a backslash path examined on a POSIX host, where `parent_path()` returns empty. Tracing where these paths actually come from, that case is not reachable: `fname` is set from `fileChooser()` or argv on the machine the program is running on, so a backslash path only ever exists in the Windows build, and there `std::filesystem::path` treats `\` as a separator anyway. The justification did not survive being checked.

So `src/tools/filepath.{h,cpp}` and `test/test_filepath.cpp` are deleted and both dialogs call `parent_path()` directly — 198 lines out, 27 in. The tests went with it; they would have been testing libc++.

`just build`, `test-all` (4/4), `check` and `build-werror` all clean. Rebased onto the master merge and pushed as bd30af21d.

tomburns:
Will merge once build is green per previous permission

PR #97, opened 2026-09-21
feat(gui): make the GUI native on macOS — system menu bar, ⌘ shortcuts, Finder opens

tomburns:
Makes the BurrTools GUI look and behave like a native macOS application, in portable C++ — **no Objective-C source files are added**; the one Cocoa call goes through the runtime's plain-C API.

The reasoning behind each decision is in the commit messages, and the findings that constrain the code are in comments beside it.

## What changes

| | Before | After |
|---|---|---|
| Menu bar | A strip inside the document window | The real macOS system menu bar |
| Menu structure | File / Toggle 3D / Export / Status / … | BurrTools / File / Puzzle / View / Window / Help |
| Shortcuts | F2 save, F3 load, no ⌘ anywhere | ⌘N ⌘O ⌘S ⇧⌘S ⌘W ⌘I ⌘3 ⌘? ⌘, ⌘Q — F-keys still work |
| Widgets | FLTK's default Motif-derived boxes | The `oxy` scheme (all platforms) |
| Opening files | Menu only | Finder double-click, Dock drop, Open With |
| Window title | `BurrTools - /long/path/x.xmpuzzle` | `x.xmpuzzle`, with the edited dot in the close button |
| Unsaved changes | Cancel, or lose your work | Save / Don't Save / Cancel (all platforms) |
| Icons | Generic macOS application icon | Application icon, plus a distinct document icon on `.xmpuzzle` files |
| Appearance | Followed the system, which FLTK ignores | Still follows the system; see the dark-mode caveat under *Known limitations* |

There is deliberately **no Edit menu**: the only candidate item is *Edit Comment*, and BurrTools has no Undo, Cut, Copy or Paste. A one-item Edit menu reads worse than none.

Linux and Windows keep the historical menu table verbatim. The only cross-platform changes are the widget scheme and the save prompt, both intentional.

## Structure

Two focused units come out of the 134 KB `mainwindow.cpp`:

- **`src/gui/platform.{h,cpp}`** — the single place this codebase branches on operating system. Six functions, each with a portable no-op or fallback body, so callers never carry an `#ifdef`.
- **`src/gui/mainmenu.{h,cpp}`** — both menu tables, built from shared callbacks, plus the drift check that keeps them honest.

`mainwindow.cpp` gains exactly one `#ifdef __APPLE__`, at the menu bar's construction, because `LFl_Sys_Menu_Bar` cannot exist off macOS. `main.cpp` gains none.

## Bugs found and fixed during review

Four defects were caught by review rather than by a user:

- **The Window menu was destroyed permanently.** FLTK inserts it into the *live* menu array at first `show()`; rebuilding from the *static* table wiped it the first time a shape was loaded, and `window_menu_installed` kept it from ever coming back. Minimize ⌘M and the window list were gone for the rest of the process.
- **`openFromSystem()` was re-entrant.** FLTK drains its dropped-files list from *nested* `Fl::wait()` loops, so opening several files at once stacked confirmation dialogs whose answers applied to the wrong puzzle.
- **The menu-table drift check was unarmed** — it ran only under a manual command, was a self-comparison off macOS, and was absent from CI.
- **A failed menu lookup was silent and latched**, so Export could have become permanently stuck enabled with no warning.

## Testing

`test_burrtools` links `src/lib` only and has no FLTK dependency, so there is no GUI unit-test harness and this PR does not add one. Instead it adds `burrtools --self-check`, a headless mode that runs the menu-table invariant and exits, wired into `just check-gui` and into the macOS CI job — so the two-table design's one real hazard is now gated automatically.

Gate: `just build && just test-all && just check && just build-werror && just check-gui`, green from a clean rebuild (4/4 suites).

**A 21-item manual checklist was walked in full, exercised against a real `.app` bundle** rather than the bare binary, which is the only way several of them can be tested: the bare binary has no `Info.plist`, so it receives neither the document-type association nor the icons. The bundle was confirmed self-contained first — every dynamic dependency resolves to `/usr/lib` or a system framework, nothing from Homebrew.

Four things were established by evidence rather than assumption:

- **⌘Q still runs the unsaved-changes prompt.** Traced through FLTK: `applicationShouldTerminate:` → `Fl::handle(FL_CLOSE)` → `Fl::default_atclose` → virtual `hide()` → `mainWindow_c::hide()`. On cancel `win->shown()` stays true and FLTK returns `NSTerminateCancel`. This was the one blocking risk identified up front; the planned fallback proved unnecessary.
- **Finder opens fire on both the cold-launch and already-running paths**, delivered by Apple Event rather than `argv`, so `show(argc, argv)` does not also load the file.
- **The document-style title is applied.** Reading the running application's window list via `CGWindowListCopyWindowInfo` — read-only, needing no Accessibility permission — reported `Bermuda.xmpuzzle` after Launch Services delivered the document.
- **The system font is genuinely applied.** Measured in the running app, rendered text width at 13pt was identical for the remapped `FL_HELVETICA` and the system UI font, and differed from real Helvetica — so the remap reaches FLTK's rendering, not just its font table.

## Icons

`scripts/make-macos-icons.sh` builds both `.icns` files from committed artwork (`mac/icon-source.png`, `mac/icon-source-doc.png`), so they are reproducible rather than opaque binaries.

The script squares a non-square source itself, and the way it does that matters: it **crops** to the smaller dimension rather than padding out to the larger one, because `sips -c` pads with `--padColor` and that **flattens the alpha channel outright** — padding a transparent render yields a solid background. The crop is reported, naming the pixels removed, so artwork whose margins cannot absorb it shows as clipping in the script output instead of being discovered later in Finder.

The plist icon keys are emitted only when the corresponding `.icns` exists, so a bundle never names a missing icon file.

## Known limitations

Stated here rather than left to be discovered:

- **The save prompt's buttons read [Don't Save] [Save] [Cancel]**, not the Mac-conventional order. FLTK hardcodes the *middle* button of `fl_choice` as the return button, so Save cannot be both rightmost and the Enter default. Enter saves and Escape cancels; placement was traded for safety deliberately.
- **Opening several files at once loads the first**, since BurrTools shows one puzzle at a time.
- **A single Finder open delivers two events** — observed, absorbed by the re-entrancy guard, root cause still unexplained and worth a follow-up.
- **Dark mode is only skin deep.** The app no longer declares itself light-only (`NSRequiresAquaSystemAppearance` was dropped in a057c6b), so under a dark system AppKit draws the title bar and system menu bar dark — but everything FLTK draws stays light, because **FLTK 1.4.4 has no dark mode on macOS**. Its Cocoa `get_system_colors()` hardcodes the light palette at `src/drivers/Cocoa/Fl_Cocoa_Screen_Driver.cxx:215-217`; the code that would read the real system colours is immediately below under `#if 0`, commented as not running on all macOS versions. Making it genuinely work means setting `Fl::background`/`background2`/`foreground` from the detected appearance, then handling the 125 XPM icons (drawn for a light background) and the widgets with hard-coded colours. Not attempted here.
- Still not Mac-like, and out of scope: single-document only, no Open Recent, the 125 in-app XPM icons are 1× and soft on Retina, unsigned (first launch needs right-click → Open), modal dialogs rather than sheets.

## Note for reviewers

`oxy` replaced `gleam` late in the branch (3ff089b) after a side-by-side comparison. Both remap the `FL_THIN_*` boxtype variants as well as the plain boxes — which matters because `separator.cpp` draws with `FL_THIN_DOWN_BOX` and would otherwise be the one widget left in the old style. Unlike `gleam`, `oxy` also calls `Fl::scrollbar_size(15)` against FLTK's default of 16, so scrollbars are a pixel narrower.


akoehn:
> Linux and Windows keep the historical menu table verbatim. The only cross-platform changes are the widget scheme and the save prompt, both intentional.

y tho

akoehn:
> Launch Services had nine BurrTools bundles registered simultaneously — /Applications, ~/Applications, ~/Applications/Old, four under ~/Downloads, a mounted .dmg volume, and the freshly built one — every one claiming CFBundleIdentifier org.burrtools.burrtools and the .xmpuzzle type. A double-click can therefore open a different, older BurrTools and look exactly like the feature being broken. Confirm which binary actually responded (ps -eo pid,lstart,command | grep burrtools) before concluding anything, and register the bundle under test explicitly with lsregister -f "$PWD/BurrTools.app".

This sounds like a problem specific to @tomburns :upside_down_face:

tomburns:
> > Linux and Windows keep the historical menu table verbatim. The only cross-platform changes are the widget scheme and the save prompt, both intentional.
> 
> y tho

Discussed offline; similar improvements for other platforms can be made in follow up PRs, tuned for their OS idioms rather than porting over an opinionated macOS layout to all three.  Cost of variation is minor here, and worth the ergonomic wins.

tomburns:
Follow-ups for the other two platforms are now tracked separately, since this PR deliberately leaves Linux and Windows on the historical menu table:

- #100 — Windows: bring the GUI up to platform conventions (Ctrl shortcuts, exe icon, file association, titles)
- #101 — Linux: add desktop integration (.desktop, MIME type, icons) and platform-conventional shortcuts and titles

Both mirror, for their platform, what this PR does for macOS. Neither blocks this one; the `src/gui/platform.{h,cpp}` seam added here is the intended place for their implementations to hang.

grssam:
Tested the build. Looks neat. Noticed a quirk:

1. when any dialog bar appears, for example, Cmd O/F2, the fading is clipped.
image
Happens irrespective of the fact if I have any piece rendered behind or not, the entire render window goes blank and un-faded like that in thie screenshot

tomburns:
Thanks — good catch, and it turned out to be two separate bugs rather than one. Both are fixed in d6ba81dc5.

**The fade isn't ours.** FLTK presents the native file dialog as a *sheet* (`Fl_Native_File_Chooser_MAC.mm:701`, `beginSheetModalForWindow:`), and AppKit dims the window a sheet is attached to. But FLTK gives every subwindow a child `NSWindow` of its own (`Fl_cocoa.mm:733`, `addChildWindow:`), and AppKit's dim does not descend into child windows. `voxelFrame_c` is the only `Fl_Gl_Window` in the application, so the 3D view was the only thing that kept its full brightness.

**One correction to the report:** the fade wasn't clipped — it was complete everywhere except the render view. Mapping every pixel of your screenshot, the undimmed region is exactly one rectangle, x 674–3424 / y 132–2182, which is the render view. The left hand panel has *zero* undimmed pixels. So there's no second problem on the left to go looking for.

**The blank render view was a genuinely separate defect** with its own cause: while the dialog is up FLTK just sits in `Fl::wait()`, and nothing damages the GL child window, so it never repainted. `fileChooser()` now drives the redraw for the dialog's duration, which puts the puzzle back on screen — dimmed along with everything else — instead of leaving it empty. That's why it happened "irrespective of whether I have any piece rendered behind".

**On matching the dim.** AppKit exposes no API for the wash it draws, so it had to be measured. I built a probe window painting four known grey levels twice — once with FLTK widgets, once in an `Fl_Gl_Window`, which the dim doesn't reach — and screenshotted it with a sheet attached:

```
in    255  216  128   64        ->  out = 0.385 * in + 13.2
out   111   97   62   38            (grey 21.4/255 at alpha 0.615)
```

Reproduces all four levels to better than one part in 255. Those constants also predict the 216 → 97 in *your* screenshot, on your machine — so it's a property of the compositor, not of one display profile. Across the seam the render view now reads 96 against the panel's 97, where it read 216.

Two ordering constraints are worth knowing about if you're reading the diff: the wash goes on *after* `PostDraw()`, because that's where image export reads the frame buffer back (`imageExport_c::PostDraw` → `trEndTile` → `glReadPixels`) and exports must not carry it; and it's skipped while picking, because `pickShape()` reruns `draw()` under `glRenderMode(GL_SELECT)` and a quad covering the whole view would be the nearest hit in every pick.

Verified locally on macOS 26.6.2 with `just build && just test-all && just check && just build-werror && just check-gui`, all green. CI hasn't run on the new commit yet. Would appreciate a re-test on your machine — and if you had further items after the "1.", I'd like to hear them.

tomburns:
Before:
Screenshot 2026-09-22 at 6 43 58 AM


After:
image

tomburns:
More After:

Screenshot 2026-09-22 at 6 47 46 AM

Screenshot 2026-09-22 at 6 48 34 AM

grssam (approved):

grssam:
Tested on macos and windows. works as advertised. the close dialog change is on both OSes while any other change is only on macos

PR #98, opened 2026-09-21
refactor(concurrency): C++20 cooperative token budget, std::jthread migration, and solver integrity

akoehn:
## Summary

This PR modernizes the BurrTools concurrency architecture to C++20:
1. **Cooperative $N$-Token Concurrency Budget**: Enforces the invariant $\text{Active Assemblers} + \text{Active Disassemblers} \le N$ via a cooperative handoff protocol in `disassemblerPool_c` seeded to $N$ worker permits, keeping CPU cores saturated without oversubscription.
2. **Single Integration Point (Zero $4\times$ Duplication)**: Concurrency budgeting and lifecycle management are encapsulated in `disassemblerPool_c` and `assembler_cb`, serving all 4 exact-cover engines (`assembler_0` DLX/SIMD, `assembler_1` DLX/SIMD) uniformly.
3. **Complete `std::jthread` & C++20 Modernization**: All threads across the codebase (`assembler_0`, `assembler_1`, `SimdHuangCover`, `disassemblerPool_c`, `thread_c`, and Python `SolutionIterator`) now run on `std::jthread` with `std::stop_token` and RAII joins. Condition variables in `disassemblerPool_c` use `std::condition_variable_any` with stop-token-aware waits.
4. **Thread-Safe Problem Metrics**: Converted `solveState`, `numAssemblies`, `numSolutions`, and `usedTime` in `problem_c` to `std::atomic`, eliminating all concurrent read/write data races under ThreadSanitizer.
5. **Solver Integrity & Replay Fixes**: Cleared search stacks on completion in `assembler_1_c::simdSearch`, added bounds checks in `assembler_0_c::setPosition`, corrected abort accounting and removed UB `const_cast` in `SimdHuangCover::parallelSolve`, and added architecture-gated `BURRTOOLS_NO_NEON`.
6. **Architecture Documentation**: Added `design/2026-09-21-cpp20-concurrency.md` (canonical specification) and marked `design/2026-09-19-threading-model-assessment.md` as superseded.

### Review Fixes Applied (Commit 0eb9f08)

- **Permit Seeding**: Initialized `assembler_permits = num_threads` in the constructor so `submit()` throttles the assembler only after all workers are active, restoring true pool parallelism and exercising queue backpressure.
- **Two-Stage Cancellation**: Added `requestStop()` and a `stop_requested` flag in `disassemblerPool_c`. When `solveThread_c::stopInternal()` is called, it wakes parked submitters on `cv_assembler` promptly (< 2s) while preserving in-flight tasks and reorder buffer results for `finish()` to drain.
- **Deterministic Inline Ordering**: Widened `inline_mutex` to cover sequence allocation, disassembly, and callback delivery, guaranteeing sequential solution ordering in single-threaded / inline fallback mode.
- **Atomic `simdCompleted`**: Synchronized `simdCompleted` using `std::atomic<bool>` with release/acquire semantics between solver and GUI threads (`getFinished()`).
- **Architecture-Gated SIMD Toggles**: Ifdef-guarded `BURRTOOLS_NO_AVX2` under x86 and `BURRTOOLS_NO_NEON` under ARM in `movementanalysator.cpp:simdDisabled()`.
- **Honest `thread_c::stop()`**: Reverted base `stop()` to an explicit no-op with comment clarifying that `std::jthread` was passed a callable taking no `stop_token`.

### Verification

- `just build`: Passed cleanly.
- `just test`: Fast test suite passed (3/3, ~5.3s).
- `just test-all`: Full test suite passed (4/4, including 27s stress test).
- `just check`: Static analysis with `cppcheck` passed (0 errors).
- `just test-py`: Python unit tests passed (9/9).
- `just test-regression`: 20/20 example puzzles match known-good 0.7.1 release output (both default and with `BURRTOOLS_NO_DISASM_POOL=1`).
- **Speedup Benchmark**: Multi-threaded pool achieves **3.49x speedup** over single-threaded (2.58s vs 9.01s, 558% CPU) on `SolidSixPieceBurrs.xmpuzzle`.
- **ThreadSanitizer (`build-tsan`)**:
  - `./build-tsan/test_burrtools "[pool]"`: 0 warnings (151 assertions).
  - `./build-tsan/test_burrtools "[parallel]"`: 0 warnings (127 assertions).
  - `./build-tsan/test_burrtools "[simd]"`: 0 warnings (422 assertions).

Stacked on top of #95 (`test/0.7.1-regression`).


src/lib/disassemblerpool.cpp
```diff
@@ -123,21 +124,34 @@ void disassemblerPool_c::submit(std::unique_ptr<assembly_c> a) {
...
+  // or the solve finishes / aborts.
+  cv_assembler.wait(lock, [this]() {
+    return assembler_permits > 0 ||
+           aborted.load(std::memory_order_relaxed) ||
+           finished.load(std::memory_order_relaxed);
+  });
+
+  if (assembler_permits > 0) {
+    assembler_permits--;
+  }
```
tomburns: **The cooperative token handoff serialises the whole solve pipeline to one disassembly at a time.**

`assembler_permits` is initialised to `0` (`disassemblerpool.h:104`) and the only non-error increment is the one-per-completed-task in `worker_loop` (L176). So every `submit()` parks here until a disassembly *finishes* — the permit is a completion signal, not a budget.

That is worse than a simple round-trip, because both assemblers call `getCallback()->assembly()` **while holding `callbackMutex`** (`assembler_0.cpp:1259`, `assembler_1.cpp:2585`). The parked thread is holding that mutex.

Concrete: solve `examples/PelikanBurr.xmpuzzle` with `-d` on an 8-core box. Assembler thread A finds an assembly, takes `callbackMutex`, submits, parks here. Threads B…H pile up on `callbackMutex`. The 8 pool workers have exactly one task between them. Steady state is ~1 assembler + ~1 disassembler regardless of `BURRTOOLS_THREADS`.

Two consequences worth calling out:
- this undoes the concurrent pool from #84;
- the bounded-queue backpressure path becomes dead code — `max_queue_size` is `max(64, num_threads)` but the queue can now never exceed 1, so the `test_disassembler_pool` case that submits 100 tasks "to exercise backpressure" no longer exercises anything.

I also couldn't find code matching the PR description's *"permits are permanently released to drain remaining disassembly tasks at full parallel saturation"* — `finish()` sets `finished`, which releases parked assemblers, but by then there is nothing left to submit.

Minimal fix is to seed the permit count so it acts as the in-flight budget it reads as — suggestion on the constructor below.

src/lib/disassemblerpool.cpp
```diff
@@ -62,6 +62,8 @@ disassemblerPool_c::disassemblerPool_c(
 
   // Sanity cap on worker threads to avoid resource exhaustion
   num_threads = std::min(num_threads, 256u);
+  max_queue_size = std::max<size_t>(64, num_threads);
+  max_reorder_size = std::max<size_t>(64, num_threads * 2);
```
tomburns, line 67: Fix for the throughput problem on L128–139: seed one compute permit per worker, so `submit()` throttles the assembler only once every disassembler is already busy.

```suggestion
  max_reorder_size = std::max<size_t>(64, num_threads * 2);

  // Seed one compute permit per worker thread. submit() then throttles the
  // assembler only once every disassembler is already busy, rather than
  // round-tripping through a completion for every single assembly.
  assembler_permits = num_threads;
```

With this, `assembler_permits` behaves as a counting semaphore of size `num_threads` and the queue can actually fill, which restores both the parallelism and the backpressure path the tests cover. (`std::counting_semaphore` would express it more directly, but the permit has to be waited on under `queue_mutex` together with `aborted`/`finished`, so the condvar is the right shape here.)

src/lib/solvethread.cpp
```diff
@@ -175,7 +175,7 @@ bool solveThread_c::assembly(std::unique_ptr<assembly_c> a) {
   if (parameters & PAR_DISASSM) {
     bt_assert(disasm_pool);
     disasm_pool->submit(std::move(a));
```
tomburns, line 182: **Stop no longer takes effect promptly, because this call can now block indefinitely.**

`submit()` parks on `cv_assembler` whose predicate watches only `assembler_permits`, `aborted` and `finished`. But `solveThread_c::stopInternal()` (L332–347) sets `stopPressed` and calls `assembler->stop()` — it never calls `disasm_pool->abort()`. And the worker that picked up the task tests only its own `st.stop_requested()` / `aborted`, neither of which Stop sets, so it runs the disassembly to completion first.

Scenario: user presses Stop on a puzzle whose single disassembly takes minutes. The GUI sits in `ACT_WAIT_TO_STOP` for that whole time. Before this PR `submit()` never blocked, so Stop was observed at the next assembly.

Cheapest fix is to have the stop path tell the pool, e.g. in `stopInternal()`:

```c++
  if (puzzle.getAssembler())
    puzzle.getAssembler()->stop();

  if (disasm_pool)
    disasm_pool->requestStop();   // wake parked submitters; keep queued results

  stopPressed = true;
```

`abort()` would also unblock it but discards the reorder buffer, so a distinct "stop accepting, wake waiters" entry point is probably what you want. (Flagging it here rather than in `solvethread.cpp` — that region isn't in the diff.)

src/lib/disassemblerpool.cpp
```diff
@@ -93,15 +95,14 @@ void disassemblerPool_c::check_exception() {
...
     return;
 
   if (is_inline) {
     uint64_t seq = next_submit_seq++;
     std::unique_ptr<separation_c> s;
     if (a && a->placementCount() > 1) {
+      std::lock_guard<std::mutex> lock(inline_mutex);
       s = inline_dis->disassemble(a.get());
     }
     if (on_result) {
```
tomburns, line 114: **The inline path holds `inline_mutex` only across the disassembly, not across sequence allocation or result delivery.**

`seq` is taken at L102 and `on_result` is invoked at L109, both outside the lock. With a multi-threaded assembler, two threads can interleave so that seq 5 is delivered before seq 4, and `onDisassemblyResult` runs on two threads at once.

What makes this more than theoretical: `is_inline` is exactly the `BURRTOOLS_NO_DISASM_POOL=1` configuration that AGENTS.md §4 documents as the A/B baseline, and it's also taken whenever `num_threads == 1`. So the ordered-merge guarantee the class header advertises ("preserving strict deterministic solution ordering") is dropped precisely in the configuration used as the benchmark and regression baseline — `just test-regression` output ordering becomes nondeterministic there.

Widening the lock to cover the whole inline path fixes it; the `lock_guard` then lives to the end of the `if (is_inline)` block, covering `on_result` and the `return`:

```suggestion
    std::lock_guard<std::mutex> lock(inline_mutex);
    uint64_t seq = next_submit_seq++;
    std::unique_ptr<separation_c> s;
    if (a && a->placementCount() > 1) {
      s = inline_dis->disassemble(a.get());
    }
    if (on_result) {
```

src/lib/movementanalysator.cpp
```diff
@@ -121,7 +121,10 @@ static bool disasmOptDisabled() {
 }
 
 static bool simdDisabled() {
-  return (std::getenv("BURRTOOLS_NO_DISASM_SIMD") != nullptr || std::getenv("BURRTOOLS_NO_AVX2") != nullptr || std::getenv("BURRTOOLS_NO_DISASM_OPT") != nullptr);
+  return (std::getenv("BURRTOOLS_NO_DISASM_SIMD") != nullptr ||
+          std::getenv("BURRTOOLS_NO_AVX2") != nullptr ||
+          std::getenv("BURRTOOLS_NO_NEON") != nullptr ||
+          std::getenv("BURRTOOLS_NO_DISASM_OPT") != nullptr);
```
tomburns: **`BURRTOOLS_NO_NEON` now disables the x86 kernels too.**

`simdDisabled()` is a single flag consulted by both the AVX2/AVX-512 paths (`rfw_avx2`, `rfw_avx512`) and the NEON path, so any one of these four variables turns off all of them. AGENTS.md defines `BURRTOOLS_NO_NEON` as "the ARM equivalent of `BURRTOOLS_NO_AVX2`", and `BURRTOOLS_NO_AVX2` as having "no effect on ARM".

Scenario: an x86-64 A/B run with `BURRTOOLS_NO_NEON=1` silently falls back to the scalar Roy-Floyd-Warshall closure, so the measured "NEON off" delta on x86 is really "all disassembler SIMD off" — a quietly wrong benchmark number rather than a crash.

```suggestion
  return (std::getenv("BURRTOOLS_NO_DISASM_SIMD") != nullptr ||
#if defined(__x86_64__) || defined(_M_X64)
          std::getenv("BURRTOOLS_NO_AVX2") != nullptr ||
#elif defined(__aarch64__) || defined(__ARM_NEON)
          std::getenv("BURRTOOLS_NO_NEON") != nullptr ||
#endif
          std::getenv("BURRTOOLS_NO_DISASM_OPT") != nullptr);
```

src/lib/assembler_1.cpp
```diff
@@ -2858,6 +2858,10 @@ void assembler_1_c::simdSearch(void) {
    */
   parallelInterrupted = abbort.load(std::memory_order_relaxed);
   simdCompleted = !parallelInterrupted;
+  if (simdCompleted) {
+    next_row_stack.clear();
```
tomburns, line 2869: **Newly unsynchronised mutation of `next_row_stack`, which the GUI thread reads.**

`simdSearch()` now clears `next_row_stack` / `task_stack` on the solver thread. `assembler_1_c::getFinished()` reads `next_row_stack.size()` from the GUI thread while polling progress — and that read sits *above* the `finishedMutex` guard, so it is unprotected:

```c++
  if (simdCompleted)
    return 1.0f;

  if (next_row_stack.size() == 0) return 1;   // <-- GUI thread, no lock held
  ...
  std::lock_guard<std::mutex> guard(finishedMutex);
```

The `simdCompleted` guard above it does not establish ordering: it's a plain `bool` (`assembler_1.h:137`), so there is no release/acquire pairing between the writer's `clear()` and the reader's `size()`.

Outcome is benign today — both branches return 1 — but given this PR's stated goal of a TSan-clean tree, it's a race newly introduced in the single-threaded SIMD path, which the `[parallel]` TSan tag does not cover. Making the flag atomic gives the pairing that makes the early return actually correct:

```diff
-  bool simdCompleted = false;
+  std::atomic<bool> simdCompleted{false};
```

with `simdCompleted.store(!parallelInterrupted, std::memory_order_release)` here and an acquire load in `getFinished()`. (Suggested as a diff rather than a suggestion block — `assembler_1.h:137` isn't in the diff.)

src/lib/thread.h
```diff
@@ -48,7 +48,7 @@ class thread_c {
     bool start();
 
     /** inform the thread to stop running, this is dependent on the thread */
-    virtual void stop() {};
+    virtual void stop() { t.request_stop(); }
```
tomburns: **This default `stop()` reads as working, but cannot stop anything.**

`thread_c::start()` launches the worker as `t = std::jthread([this](){ this->start_thread(); })` (`thread.cpp:53`) — a callable that takes no `std::stop_token`. `std::jthread` only passes its token to a callable whose first parameter accepts one, so nothing here ever observes the stop state and `t.request_stop()` is a no-op.

The previous `virtual void stop() {}` was at least honestly a no-op. The new body will mislead: any future `thread_c` subclass that relies on the inherited `stop()` silently never stops. Today only `solveThread_c` derives from `thread_c` and it overrides `stop()`, so nothing is broken right now — this is about the trap being set.

Either wire the token through (`start_thread(std::stop_token)`, worker polls `st.stop_requested()`), or keep it honest:

```suggestion
    /* NOTE: the callable handed to std::jthread in start() takes no
     * std::stop_token, so the thread never observes a stop request and
     * request_stop() would do nothing. Subclasses that need to be
     * stoppable must override this.
     */
    virtual void stop() {}
```

tomburns (commented):
Reviewed for correctness at high effort, focused on the concurrency changes. Six findings below, ordered roughly by severity; four carry applicable suggestions.

The headline one is **the cooperative token budget (#1)**: `assembler_permits` starts at zero and is only ever incremented per *completed* disassembly, so `submit()` blocks until a disassembly finishes — and it blocks while the assembler holds `callbackMutex`. That collapses the pipeline to ~1 assembler + ~1 disassembler regardless of thread count, which undoes the concurrent pool from #84 and leaves the bounded-queue backpressure tests exercising nothing. It's a one-line seed in the constructor to fix.

**Things I checked and found correct:**

- `assembler_0_c::setPosition`'s `pos <= piecenumber` guard is a genuine fix — the old loop read `rows[i]` out of bounds when a saved `pos` exceeded `piecenumber`.
- The replay fix in `assembler_1_c::simdSearch` and the four new replay tests: `canUseSimd()` correctly returns false when `pos != 0` or the stacks are empty, so the second `assemble()` and the post-restore `assemble()` really are no-ops in both assemblers.
- `problem_c` atomics — no copy-assignment of the class, all three constructors initialise the new members, and the `newAttrib(unsigned int)` overload exists.
- The `std::jthread` migration in `assembler_0` / `assembler_1` / `simd_huang_cover`: the `std::stop_token st = {}` default-argument trick is correct (jthread passes its own token; the direct main-thread call gets a never-stopped default), and the explicit joins precede the jthread destructors.
- Permit accounting against deadlock: every popped task increments a permit on all exit paths including the `catch` blocks and `abort()`, so there's no lost-permit hang. The problem in #1 is throughput, not liveness.
- The new `SimdHuangCover` abort test is deterministic, and the `const_cast` removal on `abort_flag` leaves no stale `const` overrides.

src/lib/disassemblerpool.cpp
akoehn: Fixed in 0eb9f08. Seeded `assembler_permits = num_threads` in the constructor. Multi-threaded speedup over single-threaded verified at 3.49x (9.01s down to 2.58s) on `SolidSixPieceBurrs`.

src/lib/disassemblerpool.cpp
akoehn: Fixed in 0eb9f08. Seeding `assembler_permits = num_threads` in the constructor restores pool parallelism and queue backpressure. On `SolidSixPieceBurrs`, wall time dropped from 9.01s (single-threaded) to 2.58s (multi-threaded), a 3.49x speedup with 558% CPU utilization.

src/lib/solvethread.cpp
akoehn: Fixed in 0eb9f08. Added `disassemblerPool_c::requestStop()` and a `stop_requested` atomic flag. When `solveThread_c::stopInternal()` is called, it signals the pool to wake any assembler thread waiting on `cv_assembler`, allowing `submit()` and the search to return promptly (< 2s) while keeping queued tasks and the reorder buffer intact for `finish()` to drain. Verified with an explicit duration check (`CHECK(stop_ms < 2000)`) in `test_disassembler_pool.cpp`.

src/lib/disassemblerpool.cpp
akoehn: Fixed in 0eb9f08. Widened `inline_mutex` to cover sequence allocation (`next_submit_seq++`), disassembly, and the `on_result` callback delivery. Verified determinism with `BURRTOOLS_NO_DISASM_POOL=1 python3 test/test_examples_regression.py` passing 20/20 against the 0.7.1 baseline.

src/lib/movementanalysator.cpp
akoehn: Fixed in 0eb9f08. Gated `BURRTOOLS_NO_AVX2` under x86 and `BURRTOOLS_NO_NEON` under ARM using architecture `#if` guards. Verified with an A/B benchmark on x86: `BURRTOOLS_NO_NEON=1` vs unset yields identical performance (0.17s vs 0.17s) with AVX2 intact.

src/lib/assembler_1.cpp
akoehn: Fixed in 0eb9f08. Converted `simdCompleted` to `std::atomic<bool>` with release store at search completion and acquire load in `getFinished()`. Verified with ThreadSanitizer running `./build-tsan/test_burrtools "[simd]"` (35 test cases, 422 assertions covering single-threaded and parallel SIMD search/replay) with 0 data races.

src/lib/thread.h
akoehn: Fixed in 0eb9f08. Reverted `thread_c::stop()` to an explicit no-op with a clarifying comment explaining that `std::jthread` was handed a callable taking no `std::stop_token`, avoiding misleading future subclasses.

akoehn:
fyi the speedup is on a 4 core plus hyper-threading system, so I consider the 3.5x speedup to be quite good.

I still want to review the code myself, do not merge yet please (did not mark as draft so we get the gh actions)

src/lib/disassemblerpool.cpp
tomburns: Verified at `0eb9f085`. `assembler_permits` is seeded to `num_threads` in the constructor, so the counter now reads as the in-flight budget it was written as: `submit()` pushes, then parks only once `num_threads` disassemblies are outstanding, and `worker_loop` returns a permit per completed task. Steady state is `num_threads` busy disassemblers rather than one, so the concurrent pool from #84 is back. The 3.49x on a 4C/8T box is consistent with that.

One detail left over from this comment: the *queue-size* half of the backpressure path is still unreachable. `work_queue.size()` is now bounded by in-flight work (`num_threads`, plus one entry per parked submitter, since the push happens before the permit is taken), while `max_queue_size` is `max(64, num_threads)` — so the `work_queue.size() < max_queue_size` clause of the `cv_producer` predicate cannot fire on any realistic core count. Flow control is now the permit counter plus the `max_reorder_size` window, which is fine; it just means the test comment about exercising a 64-entry queue no longer describes what the test does. Noted in a top-level comment rather than blocking here.

Resolving.

src/lib/disassemblerpool.cpp
tomburns: Verified at `0eb9f085` — applied as suggested, with the comment. `assembler_permits` now behaves as a counting semaphore of size `num_threads` waited on under `queue_mutex` alongside `aborted`/`finished`/`stop_requested`, and permit accounting stays non-negative on every exit path (`abort()`, the worker/merger catch blocks, and the exception path all top it back up by `num_threads`, and `submit()` only decrements when positive).

Resolving.

src/lib/solvethread.cpp
tomburns: Verified at `0eb9f085`. `requestStop()` is wired into `stopInternal()`, `stop_requested` is in the `cv_assembler` predicate so a parked submitter wakes, and `solveThread_c::assembly()` now returns false via `isStopRequested()` so the assembler unwinds instead of submitting again. Queued tasks and the reorder buffer survive for `finish()` to drain, which is the distinction from `abort()` I was after. The `CHECK(stop_ms < 2000)` in `test_disassembler_pool.cpp` pins the behaviour.

One loose end: `requestStop()` also calls `cv_producer.notify_all()`, but the `cv_producer` predicate in `submit()` tests only the queue/reorder-window bounds and `aborted` — not `stop_requested` — so that notify is a spurious wake and nothing more. A submitter parked on the reorder-window bound at Stop time isn't released by `requestStop()`; it unblocks only as the merger drains, so it's a latency wrinkle rather than a hang. Adding `|| stop_requested.load(...)` to that predicate (and returning early after the wait, as the `aborted` check already does) would make the notify mean what it looks like it means.

Resolving — the reported problem is fixed; the predicate nit is in a top-level comment.

src/lib/disassemblerpool.cpp
tomburns: Verified at `0eb9f085` — `inline_mutex` now covers sequence allocation, the disassembly and the `on_result` delivery, and the lock ordering stays consistent (`callbackMutex` → `inline_mutex` → solutions lock), so no new deadlock edge.

Correction to my original comment, though: I overstated the reachability. Both assemblers call `getCallback()->assembly()` while holding `callbackMutex` (`assembler_0.cpp:1257`, `assembler_1.cpp:2585`), which already serialises every production call into `submit()` — so two assembler threads could not in fact have interleaved seq allocation and delivery here. The widened lock is still the right shape, since it makes the ordering guarantee a property of the class rather than of its one caller, but it was not fixing a live bug.

Resolving.

src/lib/movementanalysator.cpp
tomburns: Verified at `0eb9f085` — applied as suggested, and the guards match the ones already used for the kernels themselves (`movementanalysator.cpp:133` for the AVX2/AVX-512 block, `:196` for NEON, and the dispatch at `:298`/`:324`), so `BURRTOOLS_NO_NEON` on x86 and `BURRTOOLS_NO_AVX2` on ARM are now inert as AGENTS.md documents them.

Resolving.

src/lib/assembler_1.cpp
tomburns: Verified at `0eb9f085` — `simdCompleted` is `std::atomic<bool>`, stored with release at search completion and loaded with acquire in `getFinished()`. That closes the race I actually cared about: the plain `bool` written on the solver thread and read on the GUI thread.

For the record, on the `next_row_stack` half my suggestion was incomplete, and I'd rather say so than leave it looking fully closed. Acquire/release only helps the reader that observes `true`, and that reader returns early without touching the vector. A reader that observes `false` and then calls `next_row_stack.size()` is still unordered against the `clear()` — no placement of the flag store fixes that; only taking `finishedMutex` around the clear and moving the `size()` read under the existing guard would. But that same unsynchronised read races against every `push_back`/`pop_back` in `iterative()` (`assembler_1.cpp:1540`–`1819`), so it is a pre-existing property of how progress polling works in this codebase, not something this PR introduced. Out of scope here.

Worth trimming one claim in the design doc, though: §7.1(5) of `design/2026-09-21-cpp20-concurrency.md` says the atomic synchronises "access to `next_row_stack` without data races under ThreadSanitizer", which is stronger than what the code does.

Resolving.

src/lib/thread.h
tomburns: Verified at `0eb9f085` — applied as suggested. `stop()` is an honest no-op again, with the comment explaining why `request_stop()` would not have worked (the callable handed to `std::jthread` in `thread.cpp:53` takes no `std::stop_token`), so the next subclass author gets told rather than trapped.

Resolving.

tomburns:
Re-reviewed `0eb9f085` against the seven threads from the first pass — all seven are addressed, and I've replied in each and resolved them. Nothing here blocks your own read-through; noting the leftovers in one place so they don't hide inside resolved threads.

My verification was code-reading plus the green CI on this head (Clang x86-64, TSan, Linux/macOS/Windows, coverage). I did not rebuild locally, so the 3.49x and the `< 2s` stop timing are your measurements, not independently reproduced — both are consistent with what the code now does.

### Three things worth a follow-up commit, none of them blocking

**1. `requestStop()` notifies `cv_producer`, but nothing in that predicate watches `stop_requested`.**

`submit()`'s producer wait tests only the queue/reorder-window bounds and `aborted`, so the `cv_producer.notify_all()` in `requestStop()` is a spurious wake and nothing more. It matters only for a submitter parked on the reorder window (`next_submit_seq - next_merge_seq >= max_reorder_size`) at Stop time: it isn't released by Stop, and unblocks only as the merger drains. Latency wrinkle, not a hang. Adding `|| stop_requested.load(std::memory_order_relaxed)` to that predicate, with an early return after the wait like the `aborted` check already has, makes the existing notify mean what it looks like it means.

**2. The queue-size backpressure clause is now unreachable, and the test comment claims otherwise.**

With the permit budget seeded to `num_threads`, `work_queue.size()` is bounded by in-flight work — `num_threads`, plus one entry per parked submitter, since the push happens before the permit is taken. `max_queue_size` is `max(64, num_threads)`, so `work_queue.size() < max_queue_size` cannot fail on any realistic core count. That's fine as design: flow control is the permit counter plus the `max_reorder_size` window, and both do work. But `test_disassembler_pool.cpp:161` still says *"Submit 100 assemblies (greater than MAX_QUEUE_SIZE = 64) to exercise backpressure"*, which is no longer what that loop exercises — it exercises the permit throttle and sequence ordering. Either retitle the comment, or drop `max_queue_size` in favour of the two bounds that are load-bearing.

**3. Two design-doc claims are stronger than the code.**

In `design/2026-09-21-cpp20-concurrency.md`:

- §7.1(5) says the atomic `simdCompleted` synchronises "access to `next_row_stack` without data races under ThreadSanitizer". It closes the race on the flag itself, which was the real fix; the unsynchronised `next_row_stack.size()` read in `getFinished()` is still unordered against the `clear()` for a reader that observes `false`. That read is pre-existing and races far more widely against `iterative()` anyway, so the code is fine as-is — it's the sentence that overreaches.
- §7.1(2) says the budget ensures `Active Assemblers + Active Disassemblers <= N`. What the permits actually bound is in-flight *disassemblies* at `N`; the assembler threads keep running alongside them, and a parked submitter has already pushed its task. "In-flight disassemblies <= N" would be the accurate claim.

Also, the doc has `file:///home/arne/development/burr-tools/...` links throughout §7.1 — worth making those repo-relative while you're in there, since they only resolve on your machine.

### One correction to my own first pass

On the inline-path thread (`disassemblerpool.cpp:104`) I said two assembler threads could interleave sequence allocation and result delivery. That was overstated: both assemblers call `getCallback()->assembly()` under `callbackMutex` (`assembler_0.cpp:1257`, `assembler_1.cpp:2585`), which already serialises every production call into `submit()`. The widened `inline_mutex` is still the right shape — it makes the ordering guarantee a property of the class rather than of its single caller — but it wasn't fixing a live bug, and the `BURRTOOLS_NO_DISASM_POOL=1` regression baseline was never actually nondeterministic.

tomburns (commented):
Fresh review pass on the latest push (commit ab7ebc1). Four issues found, ranked by severity; details inline. One earlier candidate (burrTxt.cpp ignoring submit()'s new bool return) was investigated and refuted — the callback-contract asymmetry is real but unreachable in that call site, so it's omitted here.

src/lib/disassemblerpool.cpp
```diff
@@ -107,37 +125,73 @@ void disassemblerPool_c::submit(std::unique_ptr<assembly_c> a) {
...
+  while (!hasSpace() && !terminal()) {
+    if (hadToken)
+      budget_->release();
+    cv_producer.wait(lock, [&]() { return hasSpace() || terminal(); });
+    if (terminal()) {
+      salvageAssembly(salvaged_, std::move(a));
+      return false;
+    }
+    if (hadToken &&
+        !budget_->acquire([&]() { return terminal(); })) {
```
tomburns: **Possible deadlock: `submit()` can block on `budget_->acquire()` while still holding `queue_mutex`.**

`lock` here is the `std::unique_lock<std::mutex> lock(queue_mutex)` taken at the top of `submit()`. `cv_producer.wait(lock, ...)` releases `queue_mutex` while parked and reacquires it before returning — so by the time this `budget_->acquire(...)` runs, `queue_mutex` is held again. If `acquire()` blocks (no free tokens), this thread now sits on a blocking wait *while holding* `queue_mutex`.

Scenario: every budget token is held by assembler workers. One of them enters `submit()` with a full disassembly queue, releases its token, waits on `cv_producer`, reacquires `queue_mutex`, then blocks here trying to get a token back — with `queue_mutex` still locked. Any other token holder that also calls `submit()` now blocks at `std::unique_lock<std::mutex> lock(queue_mutex)` at the top of the function and can never reach the code path that would release its own token. First thread waits for a token only the second can supply; second waits for a mutex only the first can release. Deadlock.

The comment right above this block (`// Wait for queue space WITHOUT holding our search token...`) shows the intent was exactly to avoid this class of problem — the `budget_->release()`/`acquire()` pairing just doesn't quite close the loop with `queue_mutex` held across the reacquire.

Compare with `AssemblyTaskPool::pop_task` (`assembler_pool.h`), which explicitly drops its own lock before parking on the same `ThreadBudget`.

src/lib/disassemblerpool.cpp
```diff
@@ -283,29 +397,78 @@ void disassemblerPool_c::finish() {
   check_exception();
 }
 
+void disassemblerPool_c::requestStop() {
```
tomburns: **`requestStop()` doesn't wake `cv_worker`, so an idle-parked worker won't observe it on its own.**

This function notifies `cv_producer` (line 423) and `cv_merger` (line 429), and pokes the budget (line 432), but never `cv_worker`. `worker_loop()`'s wait predicate only checks `!work_queue.empty() || finished || aborted` — it doesn't check `stop_requested`, and there's no separate mechanism tying `st` (the worker's own jthread stop token) to this flag.

Today every call site is safe: `test_disassembler_pool.cpp` and `solvethread.cpp`'s `stopInternal()` both call `requestStop()` immediately followed by `finish()`/`abort()`, either of which does notify `cv_worker`. But that makes it an unenforced contract on a new primitive — a future caller that calls `requestStop()` alone (e.g. to drain/salvage without immediately tearing the pool down) would leave any worker that's currently idle-parked stuck until an unrelated later call happens to wake it. Worth either notifying `cv_worker` here too, or documenting the "must be followed by finish()/abort()" requirement explicitly on the declaration.

src/lib/simd_huang_cover.cpp
```diff
@@ -627,52 +620,76 @@ template <typename BitsetType>
...
+  if (tasks.empty() || stop.stop_requested())
     return;
 
-  std::atomic<size_t> next_task_idx{0};
+  // Non-terminating pool (see assembler_pool.h): a worker that finds the
+  // queue empty waits for quiescence, so siblings blocked in
+  // disassemblerPool_c::submit() still have a live crew searching when they
+  // wake. No dynamic splitting here -- Huang subtrees run as seeded, skew
+  // absorbed at task granularity.
+  AssemblyTaskPool<SubtreeTask> pool;
```
tomburns: **SIMD/Huang parallel solve can't resume in-session — it silently restarts from scratch on continue.**

This `AssemblyTaskPool<SubtreeTask> pool` is entirely local to `parallelSolve()`: it's seeded from `generateTasks()` at the top of the function and never escapes (no `drain()` call, no out-parameter, nothing returned). When `runStop` fires mid-search, `parallelSolve()` just returns and the pool (with whatever subtree tasks were still outstanding) is destroyed.

Compare `assembler_1_c::parallelMultiSearch`'s DLX branch, which does `parallelTasks = pool.drain();` before setting `parallelInterrupted = true` (so the member-level `parallelTasks` holds the outstanding work for next time), and the SIMD branch itself, which sets `parallelInterrupted = true` (assembler_1.cpp ~line 2640) with no equivalent drain.

Effect: pause a hard puzzle mid-solve on the SIMD/Huang engine, hit Continue, and `generateTasks()` reruns from zero instead of picking up the remaining subtrees — all prior progress on that run is thrown away. This also runs against the invariant stated in `design/2026-09-22-assembly-work-stealing.md` that in-session continue reuses the live pool state; there's no carve-out there for this path.

src/lib/simd_exact_cover.cpp
```diff
@@ -306,10 +297,10 @@ void SimdExactCover<BitsetType>::search(
   const BitsetType &occupied,
   SearchContext &ctx,
   SolutionCallback &callback,
-  const std::atomic<bool> &abort_flag,
+  std::stop_token stop,
```
tomburns: **`std::stop_token` taken by value and re-passed on every recursive call — atomic refcount churn on the hottest loop in the solver.**

`search()` recurses once per candidate row at every backtracking node (line 369: `search(depth + 1, ..., stop, ...)`), and each call copy-constructs/destructs a `std::stop_token`, which per the standard has shared-ownership (atomic refcounted) semantics with the originating `stop_source`. On a hard puzzle this function can be invoked billions of times, so this adds real atomic synchronization traffic to the hottest path in the solver, in place of what was presumably a relaxed atomic-bool load before this rewrite — and works against the PR's own stated goal (per the design doc) of keeping cancellation checks off the hot path.

`assembler_0_c`'s DLX worker captures its stop token once per worker and never re-copies it in the hot loop (there's an in-tree comment there explaining why); worth applying the same pattern here — take `const std::stop_token &stop` (or snapshot a plain `bool`/relaxed atomic once per task) instead of by-value recursion.

src/lib/movementanalysator.cpp
```diff
@@ -116,12 +117,8 @@ void movementAnalysator_c::prepareFill(void) {
 #include <arm_neon.h>
 #endif
 
-static bool disasmOptDisabled() {
-  return std::getenv("BURRTOOLS_NO_DISASM_OPT") != nullptr;
-}
-
 static bool simdDisabled() {
```
tomburns: **Note (doc sync, not a bug in this PR):** `BURRTOOLS_NO_DISASM_OPT` is intentionally dropped here per `design/2026-09-21-cpp20-concurrency.md:252` ("planar closure and bitboard checks are always on") — `simdDisabled()` no longer reads it, folded into `SimdConfig`. That's a reasonable simplification. Flagging only because the currently-checked-out AGENTS.md on `worktree-mac-native-gui` (not part of this PR) still lists `BURRTOOLS_NO_DISASM_OPT` as a working A/B benchmarking toggle for this exact algorithmic gain; worth a follow-up to update that table when this merges, so a benchmarking agent doesn't set the var expecting an effect and silently A/B the same code path against itself.

src/lib/simd_huang_cover.cpp
akoehn: This is already noted for future work in #111

src/lib/disassemblerpool.cpp
akoehn: Good catch on the shape — holding queue_mutex across a blocking budget wait does serialize other submitters and disassembler pickups behind it, and the pop_task comparison is exactly right. One correction on severity: it can't deadlock. The waiter got there by releasing its own token first, so at least one token is always either free or held by a progressing thread (disassemblers holding tokens are mid-compute, which is finite; assembler pop-parked threads hold nothing) — the wait always resolves, at worst after one disassembly. Fixed as suggested (unlock/relock around acquire; the loop re-checks terminal/hasSpace after re-lock, and the post-loop terminal check salvages).

src/lib/disassemblerpool.cpp
akoehn: Agreed — done both: requestStop() now also notifies cv_worker (an idle worker just re-checks empty/unfinished and re-waits, so it's a harmless single wakeup), and the declaration documents that teardown still completes via finish()/abort().

src/lib/simd_exact_cover.cpp
akoehn: Confirmed and fixed — recursive search() in both SimdExactCover and SimdHuangCover now takes const stop_token& (the entry points keep by-value tokens that outlive the recursion, so the borrow is safe). Same one-line shape in both solvers.

src/lib/movementanalysator.cpp
akoehn: Checked: this branch is clean — no NO_DISASM_OPT references in AGENTS.md, bench/, or justfile here. Only the other worktree needs the table update.

tomburns (approved):

PR #99, opened 2026-09-21
Quality of life enhancements in 3D view

grssam:
Added a few quality of life enhancements along with porting a few partial changes from ColinD's WIP PR.

* Render distance improvements - don't clip the piece/model/solutions tab when zoomed in beyond a point
* 3D View Cube (taken partially from 40a701f66cdce0859838d0b6658d5f5ccecc58a5 ) . Adjust the view cube to scale based on render window size (helpful in movements tab or when not maximized full screen). Add middle mouse based panning support
* Tools window live transform preview - overlays a translucent piece on hover of any of the transform options indicating how the piece will transform along with an arrow representing the transformation (rotation, displacement, flip, etc) as well.
* Arcball mouse movement sensitivity fix for HiDPI macos


src/gui/arcball.cpp
```diff
@@ -548,3 +586,89 @@ void method2_c::addTransform(void) const
...
+    q0 = (m01 + m10) / s;
+    q1 = 0.25f * s;
+    q2 = (m12 + m21) / s;
+  } else {
+    float s = sqrtf(1.0f + m22 - m00 - m11) * 2.0f;
+    q3 = (m01 - m10) / s;
+    q0 = (m02 + m20) / s;
+    q1 = (m12 + m21) / s;
+    q2 = 0.25f * s;
+  }
```
tomburns, line 647: **High — `setRotation()` stores the transpose of the matrix it is given, so every ViewCube snap inverts the view.**

`getRotation()` just above emits the standard rotation matrix under the same naming used here (`m[0]=m00, m[1]=m10, m[2]=m20, m[3]=m01, …`), so the inverse mapping must be `q0 = (m21 - m12)/s`, `q1 = (m02 - m20)/s`, `q2 = (m10 - m01)/s`. Every branch here has those differences the other way round: branch 1 negates the vector part (giving the conjugate), branches 2–4 negate `w` (giving −conjugate). Both are the same wrong rotation, so all four paths return `Mᵀ`.

I checked numerically: `setRotation(M)` followed by `getRotation()` gives a round-trip error of 4.13 against a transpose error of 0.00.

This is on the **default** path — `rotator` defaults to `"true"` (`configuration.cpp:121`), so `voxelframe.cpp:75` constructs `method2_c`. Out of the box, clicking a face of the new view cube (`viewCube_c::snapToPart`) or releasing a cube drag (`snapNearest`) lands on the mirror-image orientation: click "FRONT" and you are looking at the back. `arcBall_c::setRotation` copies straight into `LastRot` and is correct, so only the default rotater is affected.

```suggestion
  if (tr > 0.0f) {
    float s = sqrtf(tr + 1.0f) * 2.0f;
    q3 = 0.25f * s;
    q0 = (m21 - m12) / s;
    q1 = (m02 - m20) / s;
    q2 = (m10 - m01) / s;
  } else if ((m00 > m11) && (m00 > m22)) {
    float s = sqrtf(1.0f + m00 - m11 - m22) * 2.0f;
    q3 = (m21 - m12) / s;
    q0 = 0.25f * s;
    q1 = (m01 + m10) / s;
    q2 = (m02 + m20) / s;
  } else if (m11 > m22) {
    float s = sqrtf(1.0f + m11 - m00 - m22) * 2.0f;
    q3 = (m02 - m20) / s;
    q0 = (m01 + m10) / s;
    q1 = 0.25f * s;
    q2 = (m12 + m21) / s;
  } else {
    float s = sqrtf(1.0f + m22 - m00 - m11) * 2.0f;
    q3 = (m10 - m01) / s;
    q0 = (m02 + m20) / s;
    q1 = (m12 + m21) / s;
    q2 = 0.25f * s;
  }
```

src/gui/voxelframe.cpp
```diff
@@ -1735,28 +1778,82 @@ void voxelFrame_c::draw() {
...
 
+    if (Fl::event_button() == FL_MIDDLE_MOUSE ||
+        (Fl::event_button() == FL_LEFT_MOUSE && Fl::event_state(FL_CTRL))) {
+      panning = true;
+      panStartMouseX = Fl::event_x();
+      panStartMouseY = Fl::event_y();
+      panStartX = panX;
+      panStartY = panY;
+      return 1;
+    }
```
tomburns: **High — this returns before `do_callback()`, which kills Ctrl+click voxel editing.**

Previously Ctrl+click skipped `rotater->click()` (the guard on line 1838 already excludes it) but still fell through to `do_callback()` on line 1841. Two existing gestures depend on that callback and are now dead, since the view pans instead:

* `mainWindow_c::cb_3dClick` (`mainwindow.cpp:1425`) — Ctrl+click in the Pieces tab deletes the clicked voxel.
* `movementBrowser_c::cb_3dClick` (`movementbrowser.cpp:421`) — Ctrl **pulls** a piece (Shift pushes).

Shift and Alt still work because they fall through to the callback. Ctrl is the only one being swallowed.

Requiring Ctrl+Shift for left-button panning leaves both gestures intact — `mainwindow.cpp` tests `Fl::event_ctrl()` first, but we return before `do_callback()` so nothing fires, and the `!Fl::event_state(FL_SHIFT | FL_ALT | FL_CTRL)` guard on line 1838 already keeps the combo from rotating:

```suggestion
    if (Fl::event_button() == FL_MIDDLE_MOUSE ||
        (Fl::event_button() == FL_LEFT_MOUSE &&
         Fl::event_state(FL_CTRL) && Fl::event_state(FL_SHIFT))) {
      panning = true;
      panStartMouseX = Fl::event_x();
      panStartMouseY = Fl::event_y();
      panStartX = panX;
      panStartY = panY;
      return 1;
    }
```

Whatever combination you settle on, `FL_DRAG`/`FL_RELEASE` need to match.

src/gui/voxelframe.cpp
```diff
@@ -1735,28 +1778,82 @@ void voxelFrame_c::draw() {
     glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
   }
 
+  if (withViewCube && pickx < 0 && viewCube && w() >= 48 && h() >= 48)
+    viewCube->draw(rotater, w(), h(), pixels_per_unit());
```
tomburns: **High — the ViewCube is rendered into exported PNG images.**

`exportToVector` was correctly switched to `draw(false)`, but the raster export path does not go through it. `imageExport_c` drives tile rendering on this very same `voxelFrame_c` (`imageexport.cpp:155`/`:182`, via `view3D->getView()`), which reaches FLTK's normal `draw()` → `draw(true)`. `PreDraw()` returning true only skips the projection block; this guard still passes, so the cube, the axis triad and the FRONT/TOP labels get burned into every tile of the exported image — and `viewCube_c::draw` also overwrites the tile renderer's `glViewport`/`glScissor`.

`cb` is non-null only while an export is in progress (`imageexport.cpp:483` sets it, `:392` clears it), so it is a reliable discriminator:

```suggestion
  if (withViewCube && pickx < 0 && !cb && viewCube && w() >= 48 && h() >= 48)
    viewCube->draw(rotater, w(), h(), pixels_per_unit());
```

src/gui/viewcube.cpp
```diff
@@ -0,0 +1,987 @@
...
+    hover = h;
+    if (h == PART_NONE)
+      return ACT_NONE;
+    pressPart = h;
+    pressX = mx;
+    pressY = my;
+    dragging = false;
+    tracking = true;
+    if (!isNavPart(h) && rot)
+      rot->click((float)mx, (float)my);
```
tomburns: **Medium — the cube's own drag path did not get the HiDPI fix applied everywhere else.**

`handle()` feeds raw FLTK coordinates to `rot->click()` here (and to `drag()`/`clack()` on lines 516 and 547), but the arcball is configured in *pixels* — `voxelframe.cpp:86` constructs it with `pixel_w()`/`pixel_h()`. That is exactly the mismatch 070272225 fixed in `voxelFrame_c::handle` by multiplying by `pixels_per_unit()`.

Net effect on a 2× Retina display: dragging the **cube** rotates at roughly half speed around an off-centre pivot, while dragging the **model** in the same widget behaves correctly.

`handle()` has no scale parameter, so this needs three small edits — add `float pixelScale` to the declaration (`viewcube.h:45`) and definition (line 466), pass `pixels_per_unit()` at the call site (`voxelframe.cpp:1806`), then:

```suggestion
    if (!isNavPart(h) && rot)
      rot->click((float)mx * pixelScale, (float)my * pixelScale);
```

`drag()` and `clack()` need the same treatment. Note `mx`/`my` must stay unscaled for `hitTest`/`overlayRect`, which work in FLTK units.

src/gui/tooltabs.cpp
```diff
@@ -1165,7 +1195,24 @@ void ToolTab_3::cb_transform(long task) {
...
+      se = puzzle->getNumberOfShapes();
+    } else {
+      ss = shape;
+      se = shape+1;
+    }
+
+    for (int s = ss; s < se; s++) {
+      applyTask(puzzle->getShape(s), task);
+      puzzle->getShape(s)->initHotspot();
     }
```
tomburns, line 1221: **Medium — moving `case 26: fl_message(...); return;` into `applyTask` changed what that `return` exits.**

It used to abort `ToolTab_3::cb_transform` outright. Now it only leaves `applyTask`, so this loop carries on: `initHotspot()` runs and `do_callback()` fires, which makes `mainWindow_c::cb_TransformPiece` set `changed = true` — the document is marked dirty by an operation that explicitly did nothing. And because task 26 sits inside the `toAll` range on line 1205 (`(task >= 22) && (task <= 26)`), with "apply to all shapes" ticked the modal now pops **once per shape** instead of once.

Hoisting the guard back into `cb_transform` restores the old behaviour exactly:

```suggestion
    if (task == 26) {
      fl_message("Sorry minimizing is not (yet) implemented for the rhombic grid!");
      return;
    }

    for (int s = ss; s < se; s++) {
      applyTask(puzzle->getShape(s), task);
      puzzle->getShape(s)->initHotspot();
    }
```

`case 26` in `applyTask` (line 1194) then becomes unreachable and should be dropped — which also fixes the hover-preview problem noted on line 1379. Same change applies to `ToolTab_4::cb_transform` below.

tomburns (commented):
Review of the 3D view enhancements. The cube, the panning and the transform preview are all nice additions — the notes below are what I found going through the diff hunk by hunk, plus a numeric check of the new quaternion math and a syntax check of the six changed `src/gui/*.cpp` files against the project's real compile flags (all clean, no warnings).

**Three that I think block merge:**

1. `method2_c::setRotation()` stores the **transpose** of the matrix it is handed, so every view-cube snap lands on the mirror-image orientation. This is the default rotater, so it affects the feature out of the box.
2. The new Ctrl+left-drag pan returns before `do_callback()`, which silently kills Ctrl+click voxel deletion in the Pieces tab and Ctrl+click piece-pull in the movement browser.
3. The view cube gets rendered into exported PNGs — `exportToVector` was guarded, but the raster export path goes through the normal `draw()`.

Then two medium ones around the `case 26` move into `applyTask` (it changes which function the `return` exits, and it puts a modal dialog on a *hover* path), the HiDPI scale factor missing from the cube's own drag handler, and four smaller ones — GL state leakage out of `viewCube_c::draw`, the near/far plane ratio, `kMinSize` vs. the 48px cutoff, and an uncancelled FLTK timeout.

Suggested fixes are attached inline where the fix was unambiguous. A few things I specifically checked and found **correct**: `previewShapeIndex` bookkeeping (`clearSpaces()` resets it, and the preview is always last in `shapes`, so the `erase` cannot invalidate another index); `arcBall_c::getRotation`/`setRotation`/`resetRotation` and `method2_c::getRotation` against their respective `addTransform()`; member-initialiser order in `voxelFrame_c` (no `-Wreorder` risk for `just build-werror`); and `viewcube.cpp` correctly using `#pragma GCC diagnostic` rather than the CI-forbidden `#pragma GCC target`.

Two smaller notes not worth inline comments: `panX`/`panY` are stored in world units while `size` changes with zoom, so a panned view drifts if you zoom afterwards; and a middle-click or Ctrl+click landing on the cube region is consumed by `viewCube_c::handle` (which ignores `Fl::event_button()`), so a pan cannot be started there and a right-click on the cube snaps the view on release.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

src/gui/tooltabs.cpp
```diff
@@ -1306,7 +1338,24 @@ void ToolTab_4::cb_transform(long task) {
...
```
tomburns, line 1368: Same issue as `ToolTab_3::cb_transform` above — `case 26` on line 1338 now returns from `applyTask` rather than from `cb_transform`, so the shape is marked dirty and the modal repeats once per shape under `toAll`.

```suggestion
    if (task == 26) {
      fl_message("Sorry minimizing is not (yet) implemented for the rhombic grid!");
      return;
    }

    for (int s = ss; s < se; s++) {
      applyTask(puzzle->getShape(s), task);
      puzzle->getShape(s)->initHotspot();
    }
```

src/gui/tooltabs.cpp
```diff
@@ -1317,17 +1366,52 @@ void ToolTab_4::cb_transform(long task) {
...
+void ToolTab::previewTransform(long task, bool on) {
+  ToolTabContainer *c = dynamic_cast<ToolTabContainer*>(parent());
+  if (!c)
+    return;
+  if (!on || !puzzle || shape >= puzzle->getNumberOfShapes()) {
+    c->emitPreview(0, shape);
+    return;
+  }
+  voxel_c *v = puzzle->getGridType()->getVoxel(puzzle->getShape(shape));
+  applyTask(v, task);
```
tomburns: **Medium — the new live preview calls `applyTask`, so hovering the "minimize" button on a rhombic grid pops a modal dialog.**

`ToolTab_3::applyTask` and `ToolTab_4::applyTask` now contain `case 26: fl_message(...)`. Since `previewTransform` runs on hover, merely moving the mouse over that button raises a blocking "Sorry minimizing is not (yet) implemented for the rhombic grid!" dialog — before the user has clicked anything.

Removing `case 26` from both `applyTask` overrides and hoisting the check into `cb_transform` (see the comments on lines 1212 and 1355) fixes this at the same time: the preview then falls through to no transform, which is the right behaviour for a gesture the user has not committed to.

Worth a general rule here — `applyTask` is now on a hover path, so it must stay free of anything user-visible or modal.

src/gui/viewcube.cpp
```diff
@@ -0,0 +1,987 @@
...
+  glVertex2f(x0, y1);
+  glVertex2f(x0, eaveY);
+  glVertex2f(eaveX0, eaveY);
+  glEnd();
+}
+
+void viewCube_c::draw(rotater_c * rot, int winW, int winH, float pixelScale) const {
+
+  if (!rot || winW < 16 || winH < 16)
+    return;
```
tomburns, line 801: **Low/Medium — `draw()` leaks GL state into the main scene.**

The last `glLineWidth` executed in here is `1.6f` (in `drawHouseIcon`, line 771) and it is never restored. `voxelFrame_c::draw` sets `glLineWidth(3)` only inside its `!valid()` initialisation block, so from the second frame onward the coordinate-system axes drawn at `voxelframe.cpp:755` (Pieces/Shapes tabs) render at 1.6 instead of 3. Shape edge lines escape only because their widths are baked into display lists.

`GL_BLEND` is also left enabled, `GL_CULL_FACE` left disabled, and the current colour clobbered. Since this is fixed-function GL, one push/pop covers all of it:

```suggestion
void viewCube_c::draw(rotater_c * rot, int winW, int winH, float pixelScale) const {

  if (!rot || winW < 16 || winH < 16)
    return;

  glPushAttrib(GL_ALL_ATTRIB_BITS);
```

with a matching `glPopAttrib()` before the function returns.

src/gui/voxelframe.cpp
```diff
@@ -1683,7 +1725,7 @@ void voxelFrame_c::draw() {
     }
 
     // this call has to be identical to the one in image_c::prepareOpenGlImagePart
-    gluPerspective(15, 1.0*w()/h(), size+1, 3*size+1);
+    gluPerspective(15, 1.0*w()/h(), size+1, 1000*size+1);
```
tomburns: **Low — raising only the far plane does not fully fix the clipping, and it costs a lot of depth precision.**

The camera sits at `-size*2` (line 1739), so geometry of half-extent `r` spans `z ∈ [2·size − r, 2·size + r]`:

* far `3·size+1` clipped the back when `r > size + 1` — that is the bug being fixed here, and the fix is real;
* near `size+1` clips the **front** when `r > size − 1`, i.e. in essentially the same regime. Zoomed right in (`size == 1`, near `== 2`, camera 2 units out) the front of the model is still cut off.

Meanwhile near:far goes from 3:1 to ~1000:1. The renderer leans on `glPolygonOffset` plus coincident edge lines (lines 491 and 628–667); crushing depth resolution by two and a half orders of magnitude invites z-fighting between face fill and edge lines, particularly on a 16-bit depth buffer.

Deriving both planes from the actual scene radius rather than bumping one constant would fix the clipping and keep the ratio sane. Whatever you choose has to be mirrored in `image.cpp:64`, per the comment on the line above.

src/gui/image.cpp
```diff
@@ -61,7 +61,7 @@ void image_c::prepareOpenGlImagePart(voxelFrame_c * dr) {
     trRowOrder(tr, TR_TOP_TO_BOTTOM);
 
     // this call has to be the identical one as in voxelFrame_c::draw()
-    trPerspective(tr, 15, 1.0*width/height, dr->getSize()+1, 3*dr->getSize()+1);
+    trPerspective(tr, 15, 1.0*width/height, dr->getSize()+1, 1000*dr->getSize()+1);
```
tomburns: Mirrors the change at `voxelframe.cpp:1728` — see the note there on the near plane and the ~1000:1 depth ratio. Flagging it so the two stay in lockstep, as the comment above requires.

src/gui/viewcube.cpp
```diff
@@ -0,0 +1,987 @@
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#define GL_SILENCE_DEPRECATION 1
+#include <FL/Fl.H>
+#include <FL/gl.h>
+#pragma GCC diagnostic pop
+
+static const int kSizeFraction = 4;  // widget side length is winMin/kSizeFraction
+static const int kMinSize = 90;
```
tomburns: **Low — a 90px floor makes the cube swallow small 3D views.**

`overlayRect` clamps *up* to `kMinSize` before clamping down to `winW-4`/`winH-4`, with a hard floor of 40 on line 97. `voxelFrame_c::draw` only suppresses the cube below 48px. So a 60×60 viewport gets a 56×56 cube — and since `voxelframe.cpp:1821` returns 1 for anything inside `contains()`, dragging to rotate the model becomes impossible over nearly the whole view.

This hits the dialog-sized views (image export preview, movement browser) rather than the main window.

```suggestion
static const int kMinSize = 60;
```

Raising the suppression threshold in `voxelFrame_c::draw` in step, or scaling the floor with the window, would work too — the point is that the floor and the 48px cutoff should not be able to disagree this much.

src/gui/tooltabs.cpp
```diff
@@ -1317,17 +1366,52 @@ void ToolTab_4::cb_transform(long task) {
...
+  ToolTabContainer *c = (ToolTabContainer*)v;
+  if (c->previewHandler)
+    c->previewHandler(c->previewUser, 0, c->delayedClearShape);
+}
+
+void ToolTabContainer::emitPreview(voxel_c *preview, unsigned int shapeNum) {
+  Fl::remove_timeout(previewClearTimeout, this);
+  if (!preview) {
+    delayedClearShape = shapeNum;
+    Fl::add_timeout(0.06, previewClearTimeout, this);
```
tomburns, line 1534: **Low — this timeout is never cancelled on destruction.**

`ToolTabContainer` has no destructor, and nothing calls `Fl::remove_timeout` except the next `emitPreview`. If the container is torn down inside the 60 ms window — window close, or `newGridType` below — the timeout fires on a freed `this` and dereferences `c->previewHandler`.

Narrow, but a destructor closes it:

```cpp
ToolTabContainer::~ToolTabContainer(void) {
  Fl::remove_timeout(previewClearTimeout, this);
}
```

src/gui/voxelframe.cpp
tomburns: Fixed in 006b863a — left-button panning is gone entirely, so Ctrl+click now falls through to `do_callback()` as before and both `cb_3dClick` gestures (voxel delete, piece pull) work again. `FL_DRAG`/`FL_RELEASE` are consistent with the middle-button-only gate, and the tooltip was updated to match. Resolving.

src/gui/mainwindow.cpp
```diff
@@ -2173,9 +2194,13 @@ void mainWindow_c::activateShape(unsigned int number) {
 
 void mainWindow_c::activateProblem(unsigned int prob) {
 
-  if (prob < puzzle->getNumberOfProblems())
+  if (prob < puzzle->getNumberOfProblems()) {
     View3D->getView()->showProblem(puzzle.get(), prob, shapeAssignmentSelector->getSelection());
-  else
+    // only auto-fit while the user hasn't manually zoomed the Problems tab;
+    // a manual zoom (ViewSizes[1] >= 0) always wins, matching cb_TaskSelectionTab
+    if (ViewSizes[1] < 0)
```
tomburns: **Medium — this guard does not mean what the comment above it says.**

`ViewSizes[]` is only ever written when *leaving* a tab — `mainwindow.cpp:289`, `:304`, `:320`, all of the form `ViewSizes[currentTab] = View3D->getZoom()` — plus the reset to `-1` at `:4069`. Nothing in `cb_slider` or `applyWheelZoom` touches it. So `ViewSizes[1]` does not track "has the user zoomed the Problems tab"; it tracks "has the user ever switched away from the Problems tab since this file was loaded". That breaks the intent in both directions:

* **A manual zoom does not win.** Zoom with the wheel or slider while sitting on the Problems tab and `ViewSizes[1]` stays `-1`. `activateProblem()` has ~17 call sites (`:466`, `:511`, `:530`, `:547`, `:565`, `:620`, `:653`, `:663`, `:674`, `:699`, `:729`, `:759`, `:788`, `:813`, `:853`, `:884`), and they fire on ordinary problem editing — adding a shape, changing a piece count, changing the result shape. Every one of those refits and throws the zoom away. Building up a problem means fighting the camera on each edit.
* **Then it stops working entirely.** Switch to Pieces or Solutions once and `ViewSizes[1] >= 0` for the rest of the file's lifetime. Auto-fit never runs again, even when the problem's content changes completely — which is exactly when it would be useful.

A dedicated `bool problemZoomTouched` set in `LView3dGroup::cb_slider()` and `applyWheelZoom()` would give the comment's stated semantics. Reusing `ViewSizes` cannot, because its write points are tab transitions, not zoom changes.

tomburns (commented):
Three findings on 006b863a — the new-commit delta only. The eleven threads still open from the previous round are unchanged at this head; I verified each against the file contents rather than the diff, and resolved only the Ctrl+click one, which is genuinely fixed.

src/gui/mainwindow.cpp
```diff
@@ -2173,9 +2194,13 @@ void mainWindow_c::activateShape(unsigned int number) {
...
+      View3D->fitToContent();
```
tomburns: **Medium — calling this from inside `activateProblem` corrupts the *previous* tab's saved zoom.**

`cb_TaskSelectionTab`'s `TabProblems` branch runs in this order (`mainwindow.cpp:294-307`):

```
activateProblem(problemSelector->getSelection());   // :300  -> fitToContent() moves the slider
StatProblemInfo(...);
Big3DView();
ViewSizes[currentTab] = View3D->getZoom();          // :304  -> currentTab is still the OLD tab
if (ViewSizes[1] >= 0) View3D->setZoom(ViewSizes[1]);
currentTab = 1;                                     // :307
```

`currentTab` is not updated until `:307`, so line `:304` saves whatever the slider reads *after* `fitToContent()` has already overwritten it. Switching Pieces -> Problems for the first time therefore writes the fitted **Problems** zoom into `ViewSizes[0]`, and switching back to Pieces restores that wrong value. Before this commit `activateProblem` never touched the zoom, so `:304` captured the outgoing tab's zoom correctly.

It only bites while `ViewSizes[1] < 0`, but that is precisely the first Pieces -> Problems switch of every session.

Doing the fit in `cb_TaskSelectionTab` *after* the save/restore block, rather than inside `activateProblem`, avoids the ordering trap and also decouples the fit from the ~17 unrelated `activateProblem()` call sites.

src/gui/voxelframe.cpp
```diff
@@ -870,6 +875,148 @@ void voxelFrame_c::drawVoxelSpace() {
...
+  glLineWidth(6);
+
+  switch (previewHintKind) {
+    case 1: drawPreviewRotationArc(); break;
+    case 2: drawPreviewStraightArrow(true); break;
+    case 3: drawPreviewStraightArrow(false); break;
+    default: break;
+  }
+
+  glLineWidth(1);
```
tomburns: **Low — this restores the line width to 1, but the ambient width is 3.**

`voxelFrame_c::draw` sets `glLineWidth(3)` at `:1861`, inside the `!valid()` initialisation block, so it is established once and then relied on. The coordinate-system axes at `:755-771` call `glBegin(GL_LINES)` without setting a width of their own, so they inherit it.

After the first transform-preview hover, this leaves the width at 1 permanently: from the next frame onward the red/green/blue axes in the Pieces and Shapes tabs draw thin, and stay thin for the rest of the session even once the preview is cleared. Restoring `3` here would paper over it, but the width is an implicit global that two overlays now both stomp — this one and `viewCube_c::draw` (see the comment on `viewcube.cpp:795`, where the leak is `1.6f`). One `glPushAttrib(GL_ALL_ATTRIB_BITS)` / `glPopAttrib()` pair around each overlay would fix both, and would also cover the `GL_DEPTH_TEST` and `glColor` changes made here.

src/gui/tooltabs.cpp
```diff
@@ -1317,17 +1367,147 @@ void ToolTab_4::cb_transform(long task) {
...
+static int computeTransformGeometry(const voxel_c * origShape, int transformIdx,
+                                     float axis[3], float * angleDeg) {
+  if (!origShape || transformIdx < 0)
+    return 0;
+
+  int m[3][3];
+  int basis[3][3] = { {1,0,0}, {0,1,0}, {0,0,1} };
+  for (int c = 0; c < 3; c++) {
+    int x = basis[c][0], y = basis[c][1], z = basis[c][2];
+    origShape->transformPoint(&x, &y, &z, transformIdx);
```
tomburns: **High — this throws `assert_exception` on the sphere grid, on hover.**

`origShape` here is the shape's own voxel subclass. For `GT_SPHERES` that is `voxel_2_c`, and `voxel_2_c::transformPoint` (`src/lib/voxel_2.cpp:236`) opens with

```cpp
bt_assert(((*x+*y+*z) & 1) == 0);
```

All three basis vectors have an odd coordinate sum, so the **first** call fails. `meson.build` sets `debug=true` and never sets `b_ndebug`, and I confirmed against `build/compile_commands.json` that none of the 265 BurrTools translation units are compiled with `NDEBUG` — so `bt_assert` expands to the throwing form and `bt_te()` is `[[noreturn]]`.

I linked a probe against the built objects to check rather than read it off the source:

```
GT_SPHERES (ToolTab_2, transform 9):
  spheres basis0 -> THREW assert_exception: ((*x+*y+*z) & 1) == 0
  spheres basis1 -> THREW assert_exception: ((*x+*y+*z) & 1) == 0
  spheres basis2 -> THREW assert_exception: ((*x+*y+*z) & 1) == 0
GT_BRICKS (control, transform 1):
  bricks basis0 -> (1,0,0)
  bricks basis1 -> (0,0,1)
  bricks basis2 -> (0,-1,0)
```

Repro: open a spheres puzzle, Pieces tab, move the mouse over any rotate or mirror button. `TransformButtons::cb_Preview` -> `ToolTab::previewTransform` -> here -> throw, propagating out through `Fl::run()`. The `voxel_c * v` allocated at `:1457` leaks on that path as well.

Feeding an even-sum probe triple (for example `(1,1,0)`, `(0,1,1)`, `(1,0,1)`) and recovering the linear map from those would satisfy the grid's precondition, but see the separate note on the comment above — the affine grids need more than a different probe basis.

src/gui/tooltabs.cpp
```diff
@@ -1317,17 +1367,147 @@ void ToolTab_4::cb_transform(long task) {
 
 
 
+
+/* Derives what a grid-symmetry transform represents, purely geometrically:
+ * transformPoint() maps a point "around the origin" (no translation), so applying
```
tomburns: **Medium — the premise in this comment does not hold for `voxel_1_c`, so the hint is wrong on the triangular-prism grid.**

`voxel_1_c::transformPoint` (`src/lib/voxel_1.cpp:220`) is affine, not linear, and the offset is parity-dependent:

```cpp
double xs = 0.5 + 0.5* *x;
double ys = *y * sqrt(0.75);
if ((*x + *y) & 1) ys += sqrt(1.0/3); else ys += sqrt(1.0/12);
...
*y = (int)floor(ypn);
```

so the images of the basis vectors are not the columns of the transform's linear part. Same probe as above, using `trans = 1` — which `src/lib/tabs_1/rotmatrix.inc` line 2 gives as a clean 60 degree rotation about Z (`{0.5, -sqrt(0.75), 0, sqrt(0.75), 0.5, 0, 0, 0, 1}`):

```
GT_TRIANGULAR_PRISM (ToolTab_1, transform 1 = 60deg about Z):
  tri basis0 -> (-1,1,0)
  tri basis1 -> (-3,1,0)
  tri basis2 -> (-1,0,1)
```

That is `m = [[-1,-3,-1],[1,1,0],[0,0,1]]`: `det = 2` (so the `det > 0` rotation branch is taken), `trace = 1` -> `cosA = 0` -> reported angle **90 degrees**, and axis `(m21-m12, m02-m20, m10-m01) = (0,-1,4)`, normalized `(0,-0.24,0.97)`. The user gets a tilted arc sweeping 90 degrees where the actual operation is 60 degrees about Z — precisely the information the hint exists to convey.

The `det = 2` is the tell: a genuine linear part of a symmetry operation is orthogonal and must have `det = +/-1`. Rejecting anything else would at least make the affine grids fall back to "no hint" instead of drawing a confidently wrong one, and would also catch the sphere case if that assert is dealt with by probing rather than by a special case.

Reading the rotation directly out of the grid's own `rotationMatrices[trans]` would sidestep the whole reconstruction — that table is the linear part, exactly, for every grid type.

src/gui/mainwindow.cpp
tomburns: **Low/Medium — the fit is computed against the viewport the 3D view is about to stop having.**

`computeFitSize()` reads `w()`/`h()` for the aspect ratio that feeds `tanH`. Coming from the Pieces tab with `shapeEditorWithBig3DView == false`, `cb_TaskSelectionTab` runs `activateProblem()` at `:300` — which is where this fit happens — but does not call `Big3DView()` until `:303`. So the aspect ratio used is the small shape-editor view's, while the zoom it produces is applied to the enlarged view a moment later.

Same underlying shape as the `ViewSizes` ordering problem on the line above: the fit is being done inside `activateProblem()`, which has no idea what the caller is about to do to the layout. Moving it into `cb_TaskSelectionTab` after `Big3DView()` and after the save/restore block fixes the aspect ratio, the zoom-clobber, and the ~17-call-site over-triggering in one go.

tomburns (commented):
Three more on 006b863a, from a second pass over the same delta. The first is a hard failure on the sphere grid and the second makes the new hint actively misleading on the triangular grid; both are in `computeTransformGeometry()`. I verified them by linking a probe against the built objects rather than by reading the transforms off the source.

src/gui/tooltabs.cpp
tomburns: Retracting this one — I got it wrong, and the correction is the interesting part.

The "minimize" button (task 26) is `tooltabs.cpp:341-342`, in `SizeButtons` under the **Shape** column. It is a plain `LFlatButton_c` with `cb_SizeButtons_stub`, not a `TransformPreviewButton`, so it has no hover path at all. The only chain into the preview is `TransformPreviewButton::handle` -> `TransformButtons::cb_Preview` -> `ToolTab::previewTransform`, and `:117` is the sole call site of the latter. Enumerating which task ids actually get preview buttons:

```
TransformPreviewButton:  0-14, 27-32     <- hover preview
SizeButtons  (plain):    15, 22-26       <- click only
ToolsButtons (plain):    16-21, 40, 41   <- click only
```

Task 26 is click-only, so `previewTransform(26, ...)` never happens and hovering it cannot raise the dialog. I asserted the hover reachability without checking which widget class the button was built from.

What still holds is only the general point: `applyTask` is now on a hover path, so it should stay free of anything modal or otherwise user-visible. That is a guideline rather than a live defect, so there is nothing to fix here.

The two sibling comments are unaffected and still stand on their own terms — `case 26` in `ToolTab_3::applyTask` (`:1195`) and `ToolTab_4::applyTask` (`:1339`) means the `return` now exits `applyTask` rather than `cb_transform`, so the shape is still marked dirty by an operation that did nothing, and the modal still repeats once per shape under `toAll`. `ToolTab_0` (`:698`) and `ToolTab_1` (`:895`) keep the guard in `cb_transform` and are correct; only tabs 3 and 4 moved it.

Resolving.

src/gui/tooltabs.cpp
grssam: Fixed in 03e62ee0. The guard is hoisted into `ToolTab_3::cb_transform` ahead of the loop, and `case 26` is dropped from `ToolTab_3::applyTask` — it is gone from the file entirely now (`grep 'case 26' src/gui/tooltabs.cpp` returns nothing).

All three consequences you listed are closed by that:

- the modal fires once rather than once per shape under `toAll`, since the `return` now leaves `cb_transform` before `ss`/`se` are even computed;
- `do_callback()` no longer runs, so `mainWindow_c::cb_TransformPiece` cannot set `changed = true` for an operation that did nothing;
- with the case gone from `applyTask`, the hover-preview path cannot reach `fl_message` either — which was the concern you raised separately on the preview and then retracted on reachability grounds. The guideline behind it still holds, so `applyTask` is now free of anything modal.

src/gui/tooltabs.cpp
grssam: Fixed in 03e62ee0, same shape as `ToolTab_3::cb_transform`: guard hoisted into `ToolTab_4::cb_transform` ahead of the loop, `case 26` removed from `ToolTab_4::applyTask`.

`ToolTab_0` and `ToolTab_1` were already correct and are untouched — only tabs 3 and 4 had moved the guard.

src/gui/viewcube.cpp
grssam: Fixed in 03e62ee0 as suggested: `glPushAttrib(GL_ALL_ATTRIB_BITS)` at `viewcube.cpp:814`, matching `glPopAttrib()` at `:978`.

Both early returns (`:795`, `:808`) sit before the push and there is no `return` between push and pop, so it is balanced on every path. That covers the line width, `GL_BLEND`, `GL_CULL_FACE` and the current colour in one go; the partial hand-restores that used to trail the function (`glEnable(GL_DEPTH_TEST)` / `glDisable(GL_SCISSOR_TEST)`) were dropped as redundant, since `glPopAttrib` restores whatever the ambient state actually was rather than what the function assumed it was.

The second overlay you flagged on `voxelframe.cpp:935` had the same defect — `glLineWidth(6)` restored to `1` against an ambient `3` — and got the same treatment: `drawPreviewHint()` now wraps in `glPushAttrib(GL_ENABLE_BIT | GL_LINE_BIT | GL_CURRENT_BIT)` / `glPopAttrib()`.

src/gui/tooltabs.cpp
grssam: Fixed in 03e62ee0 — destructor added exactly as written:

```cpp
ToolTabContainer::~ToolTabContainer(void) {
  Fl::remove_timeout(previewClearTimeout, this);
}
```

One correction on scope, since it changes what the fix needs to cover: of the two teardown paths named, only window close was a use-after-free. `newGridType()` deletes the child `tt` (the `ToolTab`), not the container, and `previewClearTimeout` only dereferences container-level state — `previewHandler`, `previewUser`, `delayedClearShape` — through `this`, which survives that call. So the destructor closes the real hazard and `newGridType` needed nothing.

src/gui/voxelframe.cpp
tomburns: Fixed in 03e62ee0. Both planes are now derived from the content instead of one constant being bumped: `voxelFrame_c::getNearFar()` (`voxelframe.cpp:2106`) takes the camera distance `size*2` and subtracts/adds `computeContentRadius()*1.15`, so the near plane tracks the front of the model rather than cutting it, and the near:far ratio stays bounded instead of the ~1000:1 blowout that was threatening z-fighting against `glPolygonOffset` and the coincident edge lines. The `n < 0.1` and `f < n + 1.0` clamps keep it well-defined when zoomed right in.

Dropping `getSize()` from `voxelframe.h` is the part that makes this stick: the old `size+1, 1000*size+1` expression is no longer writable at either call site, so the two projections cannot silently drift apart again.

src/gui/image.cpp
tomburns: Fixed in 03e62ee0 — `image_c::prepareOpenGlImagePart` calls `dr->getNearFar(&nearPlane, &farPlane)` and passes those to `trPerspective`, so the export path shares `draw()`'s projection by construction rather than by a duplicated formula that a comment asks people to keep in sync.

`getSize()` being removed from the header means the old expression cannot be re-written here even by accident, which is what actually holds the lockstep.

src/gui/viewcube.cpp
tomburns: 03e62ee0 applies the `kMinSize = 60` suggestion, which does help the mid-size case: a 100x100 viewport goes from a 90px cube to a 60px one, and anything at or above 240x240 now gets `winMin/kSizeFraction` rather than the floor.

Leaving this open, though, because the specific case I cited is unchanged and my own suggested value is why. In `overlayRect` the `o.s > winW - 4` clamp already dominated at that size, so a 60x60 viewport still yields a 56px cube — identical before and after. The whole 48px-to-~70px band behaves the same as it did: `voxelFrame_c::draw` renders the cube from 48px up, `overlayRect`'s hard floor of 40 keeps it large, and `handle()` consumes any press landing inside `contains()`, so rotating the model by dragging is still effectively impossible in those views.

So the floor and the 48px cutoff still disagree, just less. Two small follow-ups:

- the new comment reads `keep well under voxelFrame_c::draw()'s 48px suppression cutoff`, but 60 is above 48, not under it — as written it states the opposite of what the constant does;
- either raise the `draw()` cutoff so the cube only appears once the viewport can actually hold it plus usable drag area (~`kMinSize * 2`), or scale the floor with the window instead of using a constant. Either one makes the two numbers agree by construction rather than by choosing a happier constant.

Low severity and it only affects the dialog-sized views, so it is fine as a follow-up rather than a blocker on this PR.

src/gui/mainwindow.cpp
tomburns: Fixed in 03e62ee0 with the dedicated flag, which now carries the semantics the comment claimed.

`problemZoomTouched` is set from `cb_View3dZoomChanged` (`mainwindow.cpp:338`), wired up through the new `LView3dGroup::setZoomChangeCallback` at `:4064`. The notify fires from exactly the two real user gestures — the slider stub `cb_View3dGroupSlider_stub` and `applyWheelZoom` — while `setZoom()` and `fitToContent()` both drive the slider directly and bypass it. That distinction is what makes the flag work: restoring a remembered zoom or running the auto-fit does not mark itself as a manual override, so it cannot latch on its own first use.

Both directions of the original problem are closed:

- a wheel or slider zoom while sitting on the Problems tab now sets the flag (guarded on `currentTab == 1`), so it survives the ordinary problem-editing `activateProblem()` calls instead of being refitted away on every edit;
- the flag no longer doubles as tab-switch history, so auto-fit is not permanently disabled for the rest of the file's lifetime by one visit to Pieces or Solutions.

The ~17-call-site over-triggering is gone as a side effect of the fit moving out of `activateProblem()` entirely. Cleared alongside the `ViewSizes[]` reset at `:4083`.

src/gui/mainwindow.cpp
tomburns: Fixed in 03e62ee0. `activateProblem()` is back to just `showProblem()`/`showNothing()` — the fit moved into `cb_TaskSelectionTab`'s `TabProblems` branch and now sits *after* the save/restore block:

```
Big3DView();
ViewSizes[currentTab] = View3D->getZoom();   // still the OLD tab, but nothing has refitted yet
if (ViewSizes[1] >= 0) View3D->setZoom(ViewSizes[1]);
else if (!problemZoomTouched) View3D->fitToContent();
currentTab = 1;
```

`currentTab` is still stale on the save line, but that no longer matters: nothing has touched the slider by then, so the value captured is genuinely the outgoing tab's zoom. The first Pieces -> Problems switch of a session no longer writes the fitted Problems zoom into `ViewSizes[0]`.

src/gui/voxelframe.cpp
tomburns: Fixed in 03e62ee0 the way suggested — `glPushAttrib(GL_ENABLE_BIT | GL_LINE_BIT | GL_CURRENT_BIT)` at `voxelframe.cpp:929` with the matching `glPopAttrib()` at the end of `drawPreviewHint()`, so the ambient `glLineWidth(3)` survives a preview hover and the coordinate-system axes stay at their normal width for the rest of the session.

I checked the mask actually covers what the overlay touches: inside the push/pop the only state calls are `glDisable`/`glEnable` (`GL_ENABLE_BIT`), `glLineWidth` (`GL_LINE_BIT`) and `glColor` (`GL_CURRENT_BIT`), and both helpers it calls — `drawPreviewRotationArc` and `drawPreviewStraightArrow` — use nothing but `glBegin`/`glVertex`/`glEnd`. No blend, polygon-offset or matrix-mode changes escape the mask, and there is no `return` between the push and the pop.

The `viewcube.cpp` half of this is handled separately by the `GL_ALL_ATTRIB_BITS` pair in that thread.

src/gui/tooltabs.cpp
tomburns: Fixed in fe19821d, and by removing the probe rather than repairing it. `computeTransformGeometry` now calls `origShape->getTransformMatrix((unsigned int)transformIdx, m)` (`tooltabs.cpp:1427`); there is no call into `transformPoint` left on the hover path, so `voxel_2_c`'s `bt_assert(((*x+*y+*z) & 1) == 0)` cannot fire regardless of what probe basis would have satisfied it.

Two things fall out of that which are worth recording:

- the `try`/`catch (const std::exception &)` and the `<exception>` include are gone, so the `voxel_c *` allocated further down is no longer leaked on a throw path — because there is no throw path.
- the `det != 1 && det != -1` rejection is gone too, and correctly so: `getTransformMatrix` returns the grid's own `rotationMatrices[trans]`, which is orthogonal by construction. The new `getTransformMatrix returns an orthogonal matrix for every transformation, on every grid` test in `test_grid_invariants.cpp` pins that, and `the transform-preview hint resolves to a rotation or mirror for every button, on every grid` covers the case this thread was about directly. CI is green on fe19821d.

src/gui/tooltabs.cpp
tomburns: Fixed in fe19821d via the route flagged at the end of this comment: the new `voxel_c::getTransformMatrix()` reads the grid's own `rotationMatrices[trans]` directly, so there is no reconstruction to be wrong. For `voxel_1_c` that is `tabs_1/rotmatrix.inc`, which gives `trans = 1` as the clean 60 degree rotation about Z instead of the `det = 2` matrix the basis probe produced, so the arc drawn now matches the operation actually performed.

The misleading premise in the comment is gone with the code it described, and `voxel_c::transformPoint`'s own docs now carry the warning (`voxel.h:384-390`) pointing callers who need the geometric rotation at `getTransformMatrix()` — which is where it belongs, since the trap is a property of `transformPoint`, not of this one caller.

Nice touch that `transformPoint on the coordinate basis does not recover a linear map on every grid` went in as a test: it pins the reason the old approach could not work, so it does not get reinvented.

src/gui/mainwindow.cpp
tomburns: Fixed in 03e62ee0. `Big3DView()` now runs before the fit rather than after it, so `computeFitSize()` reads the viewport the zoom is about to be applied to.

That holds synchronously, which is the part worth checking: `Toggle3DView()` swaps the geometry with an explicit `View3D->resize(...)` (`mainwindow.cpp:3233-3234`) rather than deferring to the next layout pass, and `LView3dGroup` does not override `resize`, so `Fl_Group::resize` propagates to the inner `voxelFrame_c` before `fitToContent()` reads `w()`/`h()`. Coming from the Pieces tab with `shapeEditorWithBig3DView == false`, the aspect ratio feeding `tanH` is the enlarged view's.

Moving the fit out of `activateProblem()` closed the zoom-clobber and the over-triggering in the same change, as the two sibling threads cover.

src/gui/voxelframe.cpp
grssam: Fixed in 03e62ee0, and via `glPushAttrib` rather than by restoring `3` here, for the reason you gave — the width is an implicit global that two overlays were both stomping.

`drawPreviewHint()` now wraps in `glPushAttrib(GL_ENABLE_BIT | GL_LINE_BIT | GL_CURRENT_BIT)` / `glPopAttrib()`, which covers the `glLineWidth(6)` along with the `GL_DEPTH_TEST` and `glColor4f` changes it also made. The early return sits before the push and there is no return between push and pop.

`viewCube_c::draw` got the same treatment (`GL_ALL_ATTRIB_BITS`, since it touches considerably more state) under the comment on `viewcube.cpp:795`.

src/gui/tooltabs.cpp
grssam: Fixed in fe19821d, though not by feeding it an even-sum probe triple — the probe approach is gone entirely, for the reason in your sibling comment about the affine grids.

`voxel_c` now exposes the linear part directly:

```cpp
virtual void getTransformMatrix(unsigned int trans, double m[9]) const = 0;
```

implemented in `voxel_0_c`/`voxel_1_c`/`voxel_2_c` by reading each grid's own `rotationMatrices[trans]` table (`voxel_3_c`/`voxel_4_c` inherit `voxel_0_c`'s, as they already do for `transformPoint`). `computeTransformGeometry()` reads that instead of calling `transformPoint` at all, so the assert cannot be reached from the hover path — there is no `transformPoint` call left in `tooltabs.cpp`. Being a pure virtual, the compiler would have caught any grid that failed to supply one.

The leak you noted on the throwing path went with it: no throw, no early exit before `emitPreview` takes ownership of `v`.

Verified rather than assumed, in `test/test_grid_invariants.cpp`:

- `getTransformMatrix` returns an orthogonal matrix (unit columns, mutually perpendicular, |det| == 1) for all 240 transformations on all five grids;
- it agrees exactly with `transformPoint` on GT_SPHERES for the nine indices `ToolTab_2` actually exposes, using an even-sum point — so it reads the same rotation the grid applies, not merely some orthogonal matrix;
- the hint resolves to a rotation or a mirror for every rotate/flip button on every grid, which is the property that was actually broken.

Worth flagging one thing I got wrong in between: the first version of this fix extracted the axis from the null space of `(M-I)` unconditionally, to cover 180 degree rotations where the antisymmetric part vanishes. A null-space vector's sign is arbitrary, so that silently broke CW-vs-CCW — both directions drew the same arrow. The antisymmetric part is now primary (it carries the sign) with the null space used only as the 180 degree fallback, and there is a regression case pinning it via the three CW/CCW button pairs.

src/gui/tooltabs.cpp
grssam: Fixed in fe19821d by taking the route you suggested at the end — reading the rotation out of the grid's own table instead of reconstructing it.

The `det = 2` tell was the decisive part of this: it confirmed the images of the basis vectors are not the columns of anything orthogonal on `voxel_1_c`, so no choice of probe basis recovers the linear part there. `voxel_c::getTransformMatrix()` returns `rotationMatrices[trans]` directly, which *is* the linear part exactly, for every grid — so the triangular grid now reports its true 60 degree rotation about Z rather than a tilted 90 degrees, and the basis-vector reconstruction (with its determinant sanity check) is deleted rather than guarded.

`voxel_c::transformPoint`'s doc comment claimed it transforms "around the origin", which is what made the reconstruction look reasonable in the first place. It now carries an explicit warning that this does not hold on grids whose integer coordinates are not a plain scaled Cartesian frame, and points callers wanting the geometric rotation at `getTransformMatrix()` instead.

The cross-grid coverage is in `test/test_grid_invariants.cpp`; the triangular grid is included in the case asserting every rotate/flip button resolves to a rotation or mirror, which it previously failed.

src/gui/mainwindow.cpp
grssam: Fixed in 03e62ee0 by the move you suggested, which as you noted closes this together with the zoom-clobber and the over-triggering.

The fit now runs inside `cb_TaskSelectionTab`'s `TabProblems` branch, after `Big3DView()`:

```cpp
Big3DView();
ViewSizes[currentTab] = View3D->getZoom();
if (ViewSizes[1] >= 0)
  View3D->setZoom(ViewSizes[1]);
else if (!problemZoomTouched)
  View3D->fitToContent();
```

so `computeFitSize()` reads `w()`/`h()` for the enlarged view it is about to be shown in, not the shape editor's smaller one. `activateProblem()` is back to just `showProblem()`/`showNothing()`, so it no longer has any opinion about zoom and its other ~17 call sites cannot trigger a fit during ordinary problem editing.

src/gui/viewcube.cpp
grssam: Both points taken, fixed in 4961355b — and the comment was the worse of the two, since it documented the constant as doing the opposite of what it does.

Rather than picking a happier constant, the two numbers now agree by construction, which was your actual ask:

```cpp
int viewCube_c::minimumHostSize(void) {
  // twice the cube's own floor, so there is at least as much viewport left over
  // to drag the model in as the cube occupies
  return kMinSize * 2;
}
```

`voxelFrame_c::draw()` gates on `viewCube_c::minimumHostSize()` instead of the literal `48`, and `kMinSize`'s comment is now just `floor for the cube itself; see minimumHostSize()`.

Your analysis also turned up a third case you did not name explicitly: `handle()` had **no size gate at all**, only `if (viewCube && pickx < 0)`. So below the draw cutoff — where the cube is deliberately not rendered — `viewCube->handle()` and `contains()` still ran, and an *invisible* cube went on swallowing presses in its corner. That is strictly worse than the visible case you described. It now uses the same gate, so the two cannot drift apart.

One correction on impact, since it changes how this should be weighed: the band is not reachable today. You attributed it to the dialog-sized views, but every `LView3dGroup` host sets a minimum well above it — `movementbrowser.cpp:472` and `placementbrowser.cpp:91` use `setMinimumSize(300, 300)`, `imageexport.cpp:768` and `stlexport.cpp:360` use `400, 400`, and the main window is larger still. At 300px `overlayRect` yields `300/4 = 75`, above the floor, so the floor never binds in any current host. `LView3dGroup::getMinSize()` does report `40x40`, but nothing instantiates it at that size.

So this removes a latent trap for a future host that forgets the minimum, rather than fixing a live bug — which matches your read that it is not a blocker.

grssam:
image

Highlights the features in this PR (except middle mouse panning support)

3D view cube
Transformation preview

and most important one - getting up close with your puzzle piece!
image

tomburns (approved):

